PR #262, opened 2013-03-08
when replacing children having children of their own, things fail

dbu:
this is a WIP with failing tests and some tentative fixes that seem to be wrong.

see https://github.com/symfony-cmf/symfony-cmf/wiki/PHPCR-ODM-replace-document for a proposal how to actually solve the problems.

This needs hacks for example in the BlockBundle\Document\ImagineBlock::setImage


ElectricMaxxx:
i will close, the branch is backuped on my own fork.

PR #396, opened 2013-12-25
[POC/RFC] Operations queue

dantleech:
This is an experiment for an operations queue which unities the seperate `scheduledInserts`, `scheduledUpdates`, `scheduledReorders`, etc. into a single queue. 

This enables persistance operations to be committed out in the order in which they were programmed.

This makes sense for move, reorder and delete operations which can have multiple entries in the stack for a given document. It makes less sense for "update" because the changeset is not recorded for each operation ("persist") but is calculated once before comitting.

Lots of test fails here and this is not necessarily the solution we need, just an experiment.


dantleech:
I understand it as we only need an operation queue for tree operations (move, delete, create, reorder, etc). The documents and their data would be handled by the changeset. So first you would run-off the operation queue, then you would apply the changeset -- having mandatory UUIDs will be of great use here.

Shall we schedule this for 2.0? Without mandatory UUIDs this would be more tricky I think, and as you mention there will be BC breaks.

dbu:
this sounds a lot like 2.0 to me, yes.

you are right about field changes. but there are quite some tricky things to keep in mind (field changes on moved documents and detect that we don't need the changes on a deleted document, for example)

dantleech:
Really need to have another crack at this. Maybe there is away that doesn't involve rewriting the UOW.

PR #493, opened 2014-05-20
allow to add uuid as id to proxies

ElectricMaxxx:
As described in:
https://gist.github.com/ElectricMaxxx/68a7d033a6357e757329
i would wish to create proxies (after calling `$dm->getReference('MyClassName', 'uuid');`) with the uuid as the id. This PR will make it possible that the request on the backend is able to handle the uuid. 
Normally `$session->getNode($id);` will fetch the node. Now there will be a check if the `$id` is an uuid and fetch the node by `$session->getNodeByIdentifier($id);` , which is able to look for a node by its uuid.

I tried to create more tests for those documents that live as proxies in the UoW, but i do not get it to solve the mocking and i do not know if it is clever to do so, cause if i would mock some implementation depending, we would lose the decoupled view on the backend.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/DocumentManager.php
+++ b/lib/Doctrine/ODM/PHPCR/DocumentManager.php
@@ -1219,44 +1219,33 @@ class DocumentManager implements ObjectManager
     }
 
     /**
-     * Return the node of the given object.
+     * Return the node of the given object or its hash.
      *
-     * This node is fetched by the objects uuid or its id meants the absolute path.
+     * This node is fetched by the objects uuid or its id means the absolute path.
      *
-     * @param object $document
+     * @param object|string $document
      *
      * @return \PHPCR\NodeInterface
      *
-     * @throws InvalidArgumentException if $document is not an object.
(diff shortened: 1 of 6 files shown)


ElectricMaxxx:
I think instead of rebasing 100 commits from master i will try to extract my changes in a new commit.

dbu:
you could squash those commits `git rebase -i HEAD~14` and then
cherry-pick that commit to a new branch you start from master.

dbu:
@ElectricMaxxx hm, seems we never wrapped this one upny chance you can clean it up?

dbu:
@ElectricMaxxx any chance we can wrap this one up?

dbu:
@ElectricMaxxx do you think we can wrap this one up for 2.0 as well?

ElectricMaxxx:
That would be awesome, yes.

I will try that tonight or tomorrow. I think it will be a little bit tricky.

dbu:
cool! i suggest the first step is to move bring your changes into a clean new branch from current master.

PR #585, opened 2014-12-22
[WIP] Field slugifier id generator

dantleech:
Just an idea of how this might work.


dbu:
@dantleech something for an upcoming phpcr-odm 2.0? or should we drop this and leave it to other layers of the application to generate ids?

PR #587, opened 2015-01-09
Fixed exception message

dantleech:
Closing bracket in exception message and delimit with `'`


lib/Doctrine/ODM/PHPCR/Mapping/MappingException.php
```diff
@@ -170,6 +170,6 @@ public static function lifecycleCallbackMethodNotFound($className, $methodName)
 
     public static function noTranslatorStrategy($className, $fieldNames)
     {
-        return new self("Document '" .$className."' does not have a translation strategy, but the fields ('".implode(', ', $fieldNames)."' have been set as translatable.");
+        return new self("Document '" .$className."' does not have a translation strategy, but the fields ('".implode('\', \'', $fieldNames)."') have been set as translatable.");
```
lsmith77: maybe we should instead use sprintf?

dantleech: That would be nicer, but it would mean refactoring the whole class to be consistent.

lsmith77: yeah .. I will open a new ticket for that

PR #589, opened 2015-01-18
Fix access to children of old versions of documents

### LINKED ISSUE #573: Accessing children of a version of a document throws PHPCRException
I have a class that has a property annotated like so...

     /**
     * @PHPCR\Children()
     */
    protected $accommodationTypes;

The class itself is marked versionable="full" but the children are not, they are simply versioned with the parent using "on parent version" copy.  The versioning does seem to work - I can see that the child nodes are copied to the frozen node when the parent is checked in.

When I call findVersionByName to get a specific parent document version and then try to access the first child node with Collection::first() I get the following exception...

"Doctrine\ODM\PHPCR\PHPCRException: Document is not managed and has no id"

Retrieving the editable document using find allows me to read the children just fine.  

I'm using PHPCR-ODM Version 1.1 and backing onto Apache Jackrabbit.

mellowplace:
fix #573 

I have added another state as suggested by @dbu - STATE_FROZEN

Tests are passing currently for the Jackrabbit transport - I have not been able to test DBAL because those tests are failing on the doctrine master due to some SQL setup issue by the looks of things.


lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -78,6 +78,12 @@ class UnitOfWork
     const STATE_MANAGED = 2;
     const STATE_REMOVED = 3;
     const STATE_DETACHED = 4;
+    /**
```
dantleech: Missing blank line here

````
const STATE...

/**
 * Used ...
````

File: lib/Doctrine/ODM/PHPCR/UnitOfWork.php [hunk 2]
```diff
@@ -799,6 +805,8 @@ private function doScheduleInsert($document, &$visited, $overrideIdGenerator = n
                 break;
             case self::STATE_DETACHED:
                 throw new InvalidArgumentException('Detached document or new document with already existing id passed to persist(): '.self::objToStr($document, $this->dm));
+            case self::STATE_FROZEN:
+            	throw new InvalidArgumentException('Versions of documents cannot be persisted: '.self::objToStr($document, $this->dm));
```
dantleech: Could refer to them as "Document versions" rather than "Versions of documents" ? Is that clearer?

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -2860,7 +2868,7 @@ public function findVersionByName($className, $id, $versionName)
 
         $hints = array('versionName' => $versionName, 'ignoreHardReferenceNotFound' => true);
         $frozenDocument = $this->getOrCreateDocument($className, $node, $hints);
-        $this->dm->detach($frozenDocument);
+        //$this->dm->detach($frozenDocument);
```
dantleech: This should either be removed or uncommented.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3029,11 +3037,22 @@ private function unregisterDocument($document)
      */
     public function registerDocument($document, $id)
     {
-        $oid = spl_object_hash($document);
+    	$oid = spl_object_hash($document);
```
dantleech: Indentation pb.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3029,11 +3037,22 @@ private function unregisterDocument($document)
...
-
+        
+        // frozen nodes need another state so they are managed but not included for updates
+        $nodeType = '';
+        try {
+        	$node = $this->session->getNode($id);
+        	// In some of the unit tests $node will be null because of how they are mocked
+        	$nodeType = $node ? $node->getPrimaryNodeType()->getName() : '';
+        } catch (PathNotFoundException $e) {
+        	// Can happen if the document is new and not yet persisted
```
dantleech: s/can/will

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
mellowplace: added

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
mellowplace: Removed

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3029,11 +3037,22 @@ private function unregisterDocument($document)
...
+        }
+        
+        $this->setDocumentState($oid, $nodeType=="nt:frozenNode" ? self::STATE_FROZEN : self::STATE_MANAGED);
```
dantleech: Better to not inline this..

````php
$documentState = $nodeType == 'nt:frozenNode' ? self::STATE_FROZEN : self::STATE_MANAGED;
$this->setDocumentState($oid, $documentState);
````

But then I think I might move all of this into a separate method .. do we need this logic in other places?

File: lib/Doctrine/ODM/PHPCR/UnitOfWork.php (hunk 3 above)
Comment (mellowplace) [inline lib/Doctrine/ODM/PHPCR/UnitOfWork.php]: (it no longer needs to be detached - rather these documents are given the STATE_FROZEN state so they are not included for updates etc)

File: tests/Doctrine/Tests/Models/Versioning/FullVersionableArticleWithChildren.php [hunk 7]
```diff
@@ -0,0 +1,24 @@
...
+ * @PHPCRODM\Document(versionable="full")
+ */
+class FullVersionableArticleWithChildren extends FullVersionableArticle
+{
+	/** @PHPCRODM\Children */
+	public $childArticles;
+	
+	public function __construct()
+	{
+		$this->childArticles = new ArrayCollection();
```
dantleech: Spaces not tabs :)

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
mellowplace: Fixed - was tab instead of spaces

tests/Doctrine/Tests/ODM/PHPCR/Functional/Versioning/VersioningTestAbstract.php
```diff
@@ -340,6 +343,43 @@ public function testDeletedVersionDoesNotExistAnymore($lastVersionName)
...
+    	$childArticle->author = 'mellowplace';
+    	$childArticle->topic  = 'children test - child';
+    	$versionableArticle->addChildArticle($childArticle);
+    	
+    	$this->dm->flush();
+    	
+    	$this->dm->checkpoint($versionableArticle);
+    	
+    	$version = $this->dm->findVersionByName(
+    			'Doctrine\Tests\Models\Versioning\FullVersionableArticleWithChildren',
```
dantleech: Don't use tabs

tests/Doctrine/Tests/ODM/PHPCR/UnitOfWorkTest.php
```diff
@@ -53,20 +54,50 @@ public function setUp()
         $cmf->setMetadataFor($this->type, $metadata);
     }
 
-    protected function createNode($id, $username)
+    protected function createNode($id, $username, $primaryType = "rep:root")
```
dantleech: Use single quotes, not double quotes `'rep:root'`

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3032,8 +3041,19 @@ public function registerDocument($document, $id)
...
         $this->identityMap[$id] = $document;
-        $this->setDocumentState($oid, self::STATE_MANAGED);
-
+        
+        // frozen nodes need another state so they are managed but not included for updates
+        $nodeType = '';
+        try {
+        	$node = $this->session->getNode($id);
+        	// In some of the unit tests $node will be null because of how they are mocked
+        	$nodeType = $node ? $node->getPrimaryNodeType()->getName() : '';
```
dbu: lets do something like this:

```
$frozen = $this->session->hasNode($id) && $this->session->getNode($id)->isNodeType('nt:frozen');
```

alternatively, you could add an optional argument to pass the state along. add a new hint to getOrCreateDocuments and then pass the value of that hint to registerDocument. then you do not need to check for node types in the registerDocument method.

dbu:
coming good! the tests relating to version management should be skipped when the phpcr implementation is jackalope-doctrine-dbal - it does not support versioning at all.

what are the children you map? depending on the cascading policy set in the node type, child nodes are versioned along with the parent node. do you always return the current children documents, or use the child nodes of the frozen node? and if the latter, do you fallback to current children, or somehow merge them? (honestly i am a bit unsure about how exactly this is really supposed to work in jackrabbit. the spec is [here](http://www.day.com/specs/jcr/2.0/3_Repository_Model.html#3.13%20Versioning%20Model) and [here](http://www.day.com/specs/jcr/2.0/15_Versioning.html) but rather implicit on this topic).

mellowplace:
I see what you are getting at...  there's a load of edge cases :-)

I *think* that the current test is a child document that is not versionable on a parent that IS.  Default behaviour of Jackrabbit "on parent version" is to copy the children into the frozen node.  I think if the child is versionable itself then that node might have its own versioned subgraph and I'm not sure what happens when it makes the frozen node in this case.

What I will have to do is extend all the test cases to make sure I am trying to retrieve and navigate versions of documents with:

1. Simple children that are unversioned
2. Children that are versioned themselves
3. Referencable children
4. Different "on parent version" behaviour maybe?

It may be I find other issues with this and we might have to talk about multiple fixes but I'll wait until I see the test results for that.

Does that approach sound sane to you?

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
mellowplace: Addressed - used your $frozen example - much neater and no exceptions for tests required

mellowplace:
Ok I'm pretty confident now that this fixes accessing child nodes on versions of documents, as long as...

1. The child node is an unversioned node that belongs to a versioned parent
2. The default JCR behaviour of "on parent version: copy" is used

What I haven't tested is what happens if the child is versioned itself and what happens with referenced nodes.

Would you be happy to accept this pull request with the current scope of the fix?

composer.json
```diff
@@ -17,7 +17,8 @@
         "phpcr/phpcr": "~2.1.1",
         "phpcr/phpcr-implementation": "~2.1.0",
         "phpcr/phpcr-utils": "~1.1",
-        "doctrine/instantiator": "~1.0.1"
+        "doctrine/instantiator": "~1.0.1",
+        "jackalope/jackalope-jackrabbit": "~1.0"
```
dbu: please remove this from the composer.json again - you indeed need to do this to develop, but we don't want to lock phpcr-odm to a specific phpcr implementation.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
dbu: @mellowplace can you do this tweak please?

tests/Doctrine/Tests/ODM/PHPCR/UnitOfWorkTest.php
```diff
@@ -124,7 +160,7 @@ public function testScheduleInsertCancelsScheduleRemove()
         $object = new UoWUser();
         $object->username = "bar";
         $object->id = '/somepath';
-
+
```
dbu: please remove this whitespace (but keep the empty line)

dbu:
i totally agree on merging this as its an improvement over the status quo.

did you not have some code before the last updates that checked if a frozen document was tried to be modified? what happens if i change a frozen document? trying to save it would result in an error from phpcr. silently ignoring it would be bad imho as it hides a problem. was i dreaming that i saw code about that?

can you please squash your commits into one?

@lsmith77 may i ask you to have a quick look into this if i am missing something important or if this looks good to merge.

lsmith77:
looks good .. though I must admit I am not such an expert with the versioning. Will first look at the issues with Doctrine DBAL and the --force thing .. will do so today

mellowplace:
@dbu So if you tried to save a frozen document and it got down to the Jackrabbit transport layer (which happened during my testing) you do get an exception out of Jackrabbit itself.

There is a test that was already in the project that actually asserts that the attempt to modify a frozen doc just silently fails `VersioningTestAbstract::testModifyVersion` - so I kept my changes inline with this.

I will address the final review comment and collapse commits today.

Thanks,

Rob.

lsmith77:
please rebase on master to get the DBAL tests to be fixed

dbu:
> There is a test that was already in the project that actually asserts
> that the attempt to modify a frozen doc just silently fails
> |VersioningTestAbstract::testModifyVersion| - so I kept my changes
> inline with this.

fair enough.

please write something once you cleaned up and squashed commits. (i
don't get a notification email unless you write a comment)

mellowplace:
@dbu I'm struggling with Git a bit in terms of getting the number of commits down - I managed to squash them down to 4 (two of which appear to be some sort of upstream branch merge).  Is this ok?

dbu:
good enough :-)

thanks a lot, merged!

PR #592, opened 2015-01-29
Added getFieldMapping method

akeemphilbert:
This method is called by doctrine-extension for ORM and Mongo ODM but for some reason it's missing in PHPCR

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
@@ -1573,10 +1573,7 @@ class ClassMetadata implements ClassMetadataInterface
      */
     public function getFieldMapping($fieldName)
     {
-        if ( ! isset($this->fieldMappings[$fieldName])) {
-            throw MappingException::mappingNotFound($this->name, $fieldName);
-        }
-        return $this->fieldMappings[$fieldName];
+        return $this->getField($fieldName);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -1561,6 +1562,24 @@ public function getFieldValue($document, $field)
...
+     * @return array The field mapping.
+     *
+     * @throws MappingException
+     */
+    public function getFieldMapping($fieldName)
+    {
+        if ( ! isset($this->fieldMappings[$fieldName])) {
+            throw MappingException::mappingNotFound($this->name, $fieldName);
+        }
+        return $this->fieldMappings[$fieldName];
```
dantleech: Would this work? From what I see, `fieldMappings` is an array of field names,e.g. `array('fieldOne', 'fieldTwo')` not an associative array of field mappings.

dantleech:
Have you tried using `getField`? I think that is the analogue of `getFieldMapping` in the PHPCR-ODM.

akeemphilbert:
You're right I'll have to update. The issue I'm having is in DoctrineExtensions it calls getFieldMapping on the ClassMetada for ORM and Mongo/ODM. That method isn't defined in Doctrine\Common\Persistence\Mapping\ClassMetadata but since the other two libs have it I was thinking to add it here for consistency. (As I type this I wonder if it is that the method should be added to the ClassMetadata interface in Common so that it's implemented consistently throughout?). I can hack up doctrineextensions with type checks but it doesn't seem like a bad idea to simply have getFieldMapping in PHPCR/ODM to make things consistent

dantleech:
I think it makes sense.. @dbu @lsmith77 what do you think? Should we deprecate `getField` ?

Would be great to be compatible with doctrine-extensions. Maybe @Ocramius can advise if the `getFieldMapping` should be implemented in Doctrine commons `ClassMetadata`

lsmith77:
sounds good to me .. but some of this stuff just has different array structures which can lead to other issues. so make sure it will not lead to other surprises.

Issue #598, opened 2015-02-27
Mapping @Depth?


dantleech:
Would it make sense to have a Depth mapping? (`PHPCR\ItemInterface#getDepth`)

````php
/**
 * @PHPCRODM\Depth()
 */
public $depth;
````

This would return an integer representing the depth of the document in the tree.

PR #599, opened 2015-02-27
Added depth mapping

dantleech:
This PR adds the possiblity to map the depth of the document/node to a property in the object instance.

Would close https://github.com/doctrine/phpcr-odm/issues/598

Documentation PR: https://github.com/doctrine/phpcr-odm/pull/599


lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -543,6 +543,10 @@ public function getOrCreateDocuments($className, $nodes, array &$hints = array()
                 $documentState[$class->parentMapping] = $this->getOrCreateProxyFromNode($node->getParent(), $locale);
             }
 
+            if ($class->depthMapping) {
+                $documentState[$class->depthMapping] = $node->getDepth();
```
dantleech: Where is the best place to test that this actually works without writing an entire new testcase?

dbu: maybe simply add it to a document used in https://github.com/doctrine/phpcr-odm/blob/master/tests/Doctrine/Tests/ODM/PHPCR/Functional/BasicCrudTest.php ?

lsmith77: we usually just test this with functional tests

dbu:
good idea and makes sense to me. the workaround right now would be to map the node and implement getDepth that way.

is there a way we can enforce that this field is readonly? also, we should update the depth after a move operation has been flushed?

lsmith77:
+1

dantleech:
Have added a functional test and it now updates the Depth after moving.

tests/Doctrine/Tests/ODM/PHPCR/Functional/BasicCrudTest.php
```diff
@@ -9,6 +9,8 @@
 use PHPCR\Util\UUIDHelper;
 
 use Doctrine\ODM\PHPCR\Mapping\Annotations as PHPCRODM;
+use PHPCR\Util\PathHelper;
```
dantleech: Not used.

dantleech:
Not sure if we can throw an Exception if the user tries to set a value on depth -- do we want to? setting a value wouldn't affect anything.

dbu:
i think we did that for other things that can not be changed, to not silently ignore a problem but throw an exception. its more developer friendly to tell them they did something stupid instead of just ignoring and let people look for bugs why the change is not persisted...

but no blocker for me, we can also merge as is. @lsmith77 wdyt?

lsmith77:
its fine for me as is .. or rather I cam worried that it could lead to complications when updating an existing entity via a form if we throw an exception if the property is overwritten with an invalid value.

dbu:
well. doing a form that allows to edit the depth would be quite wrong as well. you would make the end user think he can edit something there...

anyways, thanks dan.

Issue #601, opened 2015-03-03
Cannot map PHPCR property names on translated fields


dantleech:
For example, the following works:

````xml
<field name="shadowLocaleEnabled" property="shadow-on" type="boolean" />
````

but:

````xml
<field name="shadowLocaleEnabled" property="shadow-on" type="boolean" translated="true" />
````

does not, and I get:

````
Undefined index: shadow-on

/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/UnitOfWork.php:754
/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/UnitOfWork.php:1557
/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/UnitOfWork.php:1127
/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/UnitOfWork.php:2157
````

dbu:
so manually overriding the property name is ignored? that is not good.

PR #602, opened 2015-03-04
Fixed translated property name mapping

dantleech:
This PR fixes the bug where mapping a custom PHPCR property name to a translated field will fail.

Fixes https://github.com/doctrine/phpcr-odm/issues/601


dbu:
good job, thanks!

dantleech:
Hmm. Not sure this is reading the property back..

PR #603, opened 2015-03-04
Read translated custom property name

dantleech:
Follow up to: https://github.com/doctrine/phpcr-odm/pull/602

I wonder what else I missed, maybe "assoc" fields. Will add some better tests in this PR.

### DIFF
--- a/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/AttributeTranslationStrategyTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/AttributeTranslationStrategyTest.php
@@ -91,6 +91,7 @@ class AttributeTranslationStrategyTest extends \Doctrine\Tests\ODM\PHPCR\PHPCRFu
         $node = $this->getTestNode();
         $node->setProperty(self::propertyNameForLocale('en', 'topic'), 'English topic');
         $node->setProperty(self::propertyNameForLocale('en', 'text'), 'English text');
+        $node->setProperty(self::propertyNameForLocale('en', 'custom-property-name'), 'Custom property name');
         $node->setProperty(self::propertyNameForLocale('fr', 'topic'), 'Sujet français');
         $node->setProperty(self::propertyNameForLocale('fr', 'text'), 'Texte français');
         $node->setProperty('author', 'John Doe');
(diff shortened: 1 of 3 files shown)


dbu:
ups. maybe search the codebase for `$mapping['property']`

dantleech:
null fields are also still named after the field names and not the property names

dantleech:
ok, fixed the null fields and a few other things.

tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/AttributeTranslationStrategyTest.php
```diff
@@ -167,13 +171,20 @@ public function testTranslationNullProperties()
...
         $strategy->loadTranslation($doc, $node, $this->metadata, 'fr');
         $this->assertEquals('Un sujet intéressant', $doc->topic);
         $this->assertEquals('Lorem français', $doc->getText());
         $this->assertNull($doc->nullable);
+        $this->assertNull($doc->customPropertyName);
         $this->assertEquals(array(), $doc->getSettings());
+
+        $nullFields = $node->getProperty('phpcr_locale:fr' . AttributeTranslationStrategy::NULLFIELDS)->getValue();
+        $this->assertEquals(array(
+            'custom-property-name',
```
dantleech: Why do we store the nullfields of scalar values here? Would not the abscence of any field make it implicitly null?

dbu: the problem is when all translated fields are allowed to be null, you don't know if a "translation" in that language exists or not, i think. uwe worked on that, maybe he remembers more exactly.

dbu:
thanks! i merged, it certainly is more correct. if you find more issues, please tell.

hm, btw, was setting the property name manually completely ignored before or did it not work at all? if it was ignored, we should add a note to the changelog, as people who had a property name on translated fields will need to adjust their content.

tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/AttributeTranslationStrategyTest.php
dantleech: would it not be sufficient to have an extra single multivalue field which lists the active translations? I guess its done this way beacause of merging ? /cc @uwej711

dantleech:
it was completely ignored by the attribute translation strategy, so yeah, if people did map `property` and didn't realise it didn't work they will have broken things.

dbu:
ok, thanks for #606 - i think we still should do it. its one of those bugfixes that has a slight chance of breaking peoples code but its only going in the next minor version, not in a dot release.

Issue #605, opened 2015-03-06
QueryBuilder::setLocale has no effect for HYDRATE_DOCUMENT


uwej711:
in DocumentManager::getDocumentsByPhpcrQuery the locale of the query is ignored.

uwej711:
The locale is actually not even passed from the QueryBuilder to the Query, it is lost ...

dantleech:
No sure I understand. a PHPCR `QueryInterface` does not know anything about locales.

Why not use the PHPCR query object instead of the ODM query builder?

uwej711:
From http://doctrine-phpcr-odm.readthedocs.org/en/latest/reference/query-builder.html#querying-translated-documents

uwej711:
This suggest that you just use setLocale to get your content in a different language. But this does not work, the content is always returned in the current locale.

Background: I want to allow my editors to edit foreign language without changing the language of the rest of the UI. This works in Sonata in the edit view since there I can call doLoadTranslation before showing the content. But for the list view I need to have the query return the content in the correct language.

dbu:
i fear you mixed up the phpcr-odm query builder and the phpcr query builder, could that be? on phpcr level, there is indeed nothing with locales.

* there is `createPhpcrQueryBuilder()` which has no arguments and does not care about the language.
* there is `createPhpcrQuery` and `createPhpcrQuery` which accept a phpcr query with a specified query language, as in "SQL2" or "XPATH". this is *not* a locale.

i think what you want to do is getLocaleChooserStrategy()->setLocale('x').

btw, i think the feature you describe sounds a lot like https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/328 - if you have some inputs there, that would be highly appreciated.

uwej711:
@dbu this is exactly where i started from, but no, I do not mix ODM and PHPCR QueryBuilder. I will try to do a failing test ...

dbu:
ok, sorry. then that sounds like a bug, or a feature that we thought of but in the end did not finish to implement. glad if you can provide a functional test so we can start fixing this.

uwej711:
See failing test in #607

uwej711:
see https://github.com/valiton-forks/phpcr-odm/tree/sd_fixes_1_2_2 for a minimal fix (not really tested yet)

dbu:
can you open a WIP pull request that we can comment on?

i think we should wrap
[...]

PR #607, opened 2015-03-06
Demonstrate expected behaviour for QueryBuilder::setLocale

uwej711:
Failing test for issue #605 

I would expect setLocale on the QueryBuilder would return the found document in the specified locale

### DIFF
--- a/tests/Doctrine/Tests/ODM/PHPCR/Functional/QueryBuilderTranslationTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Functional/QueryBuilderTranslationTest.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace Doctrine\Tests\ODM\PHPCR\Functional;
+
+use Doctrine\ODM\PHPCR\DocumentRepository;
+use Doctrine\ODM\PHPCR\Translation\LocaleChooser\LocaleChooser;
+use Doctrine\Tests\Models\Translation\Article;
+use Doctrine\Tests\ODM\PHPCR\PHPCRFunctionalTestCase;
+use PHPCR\Util\PathHelper;
+use PHPCR\Util\NodeHelper;
+
+/**
+ * @group functional
+ */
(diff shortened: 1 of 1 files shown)


dantleech:
I will try and have a look at this tomorrow. This is surely [meant to work](https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterPhpcr.php) but couldn't find any existing tests, very odd.

uwej711:
Maybe adding a doLoadTranslation here would be a way to do it:
https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/Query/Query.php#L183

dbu:
or temporarily setting the locale on the locale chooser right before
executing the query, then restoring the previous once the query is executed.

dantleech:
It seems that it would be ideal to add a `$locale` argument to the `loadDocumentsByPhpcrQuery`, but that cascades then to `findMany()` (which can then pass the `locale` hint to the unit of work `getOrCreateDocuments` method).

But by this logic we should add the `$locale` argument to all of the document manager methods (including persist). If that is a good idea or not, I don't know, but I guess its out of the scope here.

The option would be to add translation aware methods, e.g. `getTranslatedDocumentsByPhpcrQuery`, `findManyTranslated`.

But I vote for @dbu s suggestion for now.. and we could create an issue to solve this in 2.0

dbu:
the only other option i see would be to remove the possiblility to specify the locale in the query builder. that would be more consistent with the rest. then we would say whatever you do, you need yourself to set the default locale if you want to change something about what the default locale is.

i think we can say that changing the default locale for a single call is out of scope of phpcr-odm, and you just need to change the default if you need this.

or we could offer some sort of wrapper-system where you can say $dm->defaultLocale('de')->find... defaultLocale() would return a dm decorator that changes the locale on the fly and changes it back.

dantleech:
Opened issue https://github.com/doctrine/phpcr-odm/issues/623 to discuss how to handle this

PR #609, opened 2015-03-11
Fixed depth inheritance

dantleech:
The depth mapping is not being inherited ...


dbu:
ups. makes me wonder if we could make this somehow more generic to make it easier to not forget such a thing?

dantleech:
I guess the details relating to mapping the fields could be moved into the ClassMetadata so that all of that stuff is in one place.

PR #614, opened 2015-03-19
fix #613 for queries with join from is a source

### LINKED ISSUE #613: Parameter 2 of TranslationStrategyInterface::alterQueryForTranslation should be of type SourceInterface
otherwise this fails when you have a query with a join.

uwej711:
see #613

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/TranslationStrategyInterface.php
+++ b/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/TranslationStrategyInterface.php
@@ -23,7 +23,7 @@ use Doctrine\ODM\PHPCR\Mapping\ClassMetadata;
 use PHPCR\NodeInterface;
 use PHPCR\Query\QOM\ConstraintInterface;
 use PHPCR\Query\QOM\QueryObjectModelFactoryInterface;
-use PHPCR\Query\QOM\SelectorInterface;
+use PHPCR\Query\QOM\SourceInterface;
 
 /**
  * Operations that a translation strategy must support.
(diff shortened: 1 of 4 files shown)


lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/TranslationStrategyInterface.php
```diff
@@ -122,7 +122,7 @@ public function getTranslatedPropertyPath($alias, $propertyName, $locale);
      */
     public function alterQueryForTranslation(
         QueryObjectModelFactoryInterface $qomf,
-        SelectorInterface &$selector,
+        SourceInterface &$selector,
```
dbu: this is a BC break if somebody has his own translation strategy. can you add a note on this to CHANGELOG.md please? i think we still need to break it - its not very likely people have their own strategy i hope, and we really should fix this.

uwej711: Done, not sure about the version ...

dbu:
changlog looks good to me.

@lsmith77 @dantleech can we live with this BC violation? its not elegant, but i think we really need to fix this, and afaik nobody is maintaining a public custom translation strategy. in a custom project, the change is simple to adapt to, it would only be a pain to support phpcr-odm 1.2 and 1.3 at the same time.

dantleech:
Yeah, technically we shouldn't do this. But I guess the chances are quite low that people use a custom translation strategy, its not configurable from the bundle from what I can see.

lsmith77:
+1 on the change as the fix for users is quite simple

dbu:
great, thanks uwe!

PR #615, opened 2015-03-19
Added preBind translation

dantleech:
Added yet another event.

Use case: We have a content serializer which serializes dynamic content data to multiple nodes, when the translation is bound we need to store the current content, and later resetore it in loadTranslation.

This event overlaps with the `preCreateTranslation` event, which is only called when the translation is created, this information could be passed in the `loadTranslation` event.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Event.php
+++ b/lib/Doctrine/ODM/PHPCR/Event.php
@@ -40,6 +40,7 @@ final class Event
 
     const postLoadTranslation = 'postLoadTranslation';
     const preCreateTranslation = 'preCreateTranslation';
+    const preUpdateTranslation = 'preUpdateTranslation';
     const preRemoveTranslation = 'preRemoveTranslation';
     const postRemoveTranslation = 'postRemoveTranslation';
 
@@ -55,6 +56,7 @@ final class Event
         self::postLoad => self::postLoad,
         self::postLoadTranslation => self::postLoadTranslation,
         self::preCreateTranslation => self::preCreateTranslation,
+        self::preUpdateTranslation => self::preUpdateTranslation,
         self::preRemoveTranslation => self::preRemoveTranslation,
         self::postRemoveTranslation => self::postRemoveTranslation,
     );
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -753,6 +753,14 @@ class UnitOfWork
                 new LifecycleEventArgs($document, $this->dm),
                 $invoke
             );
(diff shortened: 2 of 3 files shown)


lib/Doctrine/ODM/PHPCR/DocumentManager.php
```diff
@@ -357,11 +357,12 @@ public function find($className, $id)
      *      specified class. All others are treated as not found.
      * @param array       $ids       List of repository paths and/or uuids to
      *      find documents. Non-existing ids are ignored.
+     * @param string      $locale    Find documents in the given locales
      *
      * @return Collection list of documents that where found with the $ids and
      *      if specified the $className.
      */
-    public function findMany($className, array $ids)
+    public function findMany($className, array $ids, $locale = null)
```
dbu: did you mean to commit this bit?

lib/Doctrine/ODM/PHPCR/Event.php
```diff
@@ -56,6 +57,7 @@
         self::postLoadTranslation => self::postLoadTranslation,
         self::preCreateTranslation => self::preCreateTranslation,
         self::preRemoveTranslation => self::preRemoveTranslation,
+        self::postBindTranslation => self::postBindTranslation,
```
dbu: post bind or pre bind?

lib/Doctrine/ODM/PHPCR/DocumentManager.php
dantleech: nope

dbu:
i think some of your array PR slipped into this PR.

why do you call it preBindTranslation and not preLoadTranslation? might be more consistent with postLoadTranslation. or is loading and binding not the same?

dantleech:
Mainly because `preBindTranslation` is in the `doBindTranslation` method. `doBindTranslation` is called in the `loadTranslation` method, but also from `computeChangeset` and `bindTranslation`.

dbu:
is loadTranslation also raised in all those occasions, or only in the loadTranslation method?

dantleech:
Only in loadTranslation

dbu:
does that make sense, or should postLoadTranslation also happen when computing changeset or binding translation? i am afraid of ending up with very confusing and hard to understand events that happen in one place but not the other and are not evident. could we call ours preLoadTranslation, trigger it in all those occasions but maybe have the event provide information in which context it happened?

dantleech:
I guess one valid way might be to change it to `preUpdateTranslation`. Then the user can make use of both `preCreateTranslation` and `preUpdateTranslation` to get the same effect, much like `prePersist` and `preUpdate`.

dbu:
sounds good to me. "creation" of a translation means when the document is saved in the specific language for the first time, right? so the document itself could be existing, but a new language added?

dantleech:
Updated.

dbu:
thanks, this looks consistent now. do we need to update some documentation as well?

PR #616, opened 2015-03-19
Locale not being passed as hint in findMany

dantleech:
Looks like this was unintentional.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/DocumentManager.php
+++ b/lib/Doctrine/ODM/PHPCR/DocumentManager.php
@@ -389,7 +389,10 @@ class DocumentManager implements ObjectManager
         }
 
         $nodes = $this->session->getNodes($ids);
-        $hints = array('fallback' => true);
+        $hints = array(
+            'fallback' => true,
+            'locale' => $locale,
+        );
         $documents = $this->unitOfWork->getOrCreateDocuments($className, $nodes, $hints);
 
         return new ArrayCollection($documents);


lsmith77:
uhm the variable isn't defined in the method.

dantleech:
Hmm. damn, I suspected that I might have added that variable locally and forgotten about it ..

Is there a case for this feature? If so, how to implement it?

- adding a `findManyTranslations` method.
- make `findTranslation` accept an array (inconsistent with `find`).

- Add a locale to `findMany` (also inconsistent).

lsmith77:
I guess the only "clean" thing is ``findManyTranslations()`` which also prevents unwanted issues with the ObjectManager interface.

dbu:
this goes somewhat into the same discussion as with uwe on how to query with the non-default locale. i tend to think we should recommend to change the requested locale on the LocaleChooser. otherwise we end up adding extra parameters to (or adding a locale aware version of) a dozen methods. 
delegating all responsibility to the LocaleChooser sounds better to me. i see that we then have the drawback of changing state and being implicit, but imho the whole approach to multilanguage in phpcr-odm is implicit, having things work without explicit locale by default.

dantleech:
What if we provide a method which provides a clone of the current document manager but with a specific locale?

````php
$newLocalizedDm = $dm->localize('de');
$newLocalizedDm->find(null, '/path/to');

$dm->localize('de')->findMany(null,$ids);
````

Just an idea.

**Comment by dbu** [comment]
sounds like a good user interface to me, if we think the DocumentManager
should be concerned with the locale.

don't know how hard that would be to implement however. a decorator
around all public methods of the DocumentManager would be a huge
undertaking. and we still would need to figure out a way to temporarily
set the right locale.

**Comment by dantleech** [comment]
Well, initially I am thinking we simply add a `localize` method to the current document manager

````php
public funciton localize($locale)
{
    $dm = clone $this;
    $dm->setLocale($locale);

    return $locale;
}
````

As clone is not recursive I think this should just work (tm).

**Comment by dbu** [comment]
but there is no setLocale method on the DocumentManager.

it could work to do

```
$dm = clone $this;
$chooser = clone $this->localeChooserStrategy;
$chooser->setLocale($locale);
$dm->setLocalChooserStrategy($chooser);

return $dm;
```

i guess that would be ok to do. should we check whether the locale
chooser is clonable and throw an exception otherwise? there could be
custom strategies.

dantleech:
Well, I was thinking more that we would use the explicit `locale` hint on the applicable methods and avoid using the LocaleChooser. We could use reflection to set the localization on the cloned DM - if that property is not null then the hint would be used.

uwej711:
Cloning the DocumentManager only might not be enough as the UnitOfWork does all the heavy lifting and has quite a lot of state. Personally I still think passing the locale around explicitly in case you need a different locale than the one chosen by the LocaleChooser is the better solution.

dantleech:
For the record, what I am proposing would pass the locale explicitly from the document manager to the UOW.

dantleech:
Closing this, opened issue: https://github.com/doctrine/phpcr-odm/issues/623 to discuss further

PR #618, opened 2015-03-25
cleaning up phpdoc on query builder

dbu:
I noticed that we have some typos in the docs. Went over them and also cleaned other codestyle issues.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Query/Builder/AbstractNode.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Builder/AbstractNode.php
@@ -142,9 +142,6 @@ abstract class AbstractNode
         $cardinalityMap = $this->getCardinalityMap();
         $nodeType = $node->getNodeType();
 
-        $validChild = true;
-        $end = false;
-
         // if proposed child node is of an invalid type
         if (!isset($cardinalityMap[$nodeType])) {
             throw new OutOfBoundsException(sprintf(
@@ -214,6 +211,8 @@ abstract class AbstractNode
     /**
      * Return children of specified type.
      *
+     * @param string $type The type name.
+     *
      * @return AbstractNode[]
      */
     public function getChildrenOfType($type)
@@ -233,7 +232,7 @@ abstract class AbstractNode
     /**
      * Return child of node, there must be exactly one child of any type.
      *
-     * @return AbstractNode[]
+     * @return AbstractNode
      *
      * @throws OutOfBoundsException if there are more than one or none
      */
@@ -263,7 +262,9 @@ abstract class AbstractNode
      *
      * Note: This does not take inheritance into account.
      *
-     * @return AbstractNode[]
+     * @param string $type The name of the type.
+     *
+     * @return AbstractNode
      *
      * @throws OutOfBoundsException if there are more than one or none
      */
@@ -344,6 +345,9 @@ abstract class AbstractNode
     /**
      * Catch any undefined method calls and tell the developer what
      * methods can be called.
+     *
+     * @param string $methodName The requested nonexistent method.
+     * @param array  $args       The arguments that where used.
      *
      * @throws BadMethodCallException if an unknown method is called.
      */
--- a/lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterPhpcr.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterPhpcr.php
@@ -126,6 +126,8 @@ class BuilderConverterPhpcr
      *
      * @return array first element is the real alias to use, second element is
      *      the property name
+     *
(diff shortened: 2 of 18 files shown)


lib/Doctrine/ODM/PHPCR/Query/Builder/AbstractNode.php
```diff
@@ -142,9 +142,6 @@ public function addChild(AbstractNode $node)
         $cardinalityMap = $this->getCardinalityMap();
         $nodeType = $node->getNodeType();
 
-        $validChild = true;
-        $end = false;
```
dbu: unused code

lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterPhpcr.php
```diff
@@ -527,6 +530,9 @@ protected function doWalkConstraintComposite(AbstractNode $node, $method)
 
             $lPhpcrConstraint = $phpcrComposite;
         }
+        if (!$phpcrComposite) {
+            throw new \InvalidArgumentException('Composite must have at least one constraint');
+        }
```
dbu: this is an actual code change!
my IDE warned that $phpcrComposite might be undefined. i think this is the correct behaviour, as we would have gotten a fatal error before when there are no constraints.

dantleech: would it make more sense to do this earlier if count($children) == 0 ?

lib/Doctrine/ODM/PHPCR/Query/Builder/ConstraintFactory.php
```diff
@@ -127,7 +138,7 @@ public function fullTextSearch($field, $fullTextSearchExpression)
      * Same document constraint.
      *
      * <code>
-     * $qb->where()->same('/path/to/doc', 'sel_1');
+     * $qb->where()->same('/path/to/doc', 'sel_1')->end();
```
dantleech: ftr, the terminating `end()` is not necessary here

lib/Doctrine/ODM/PHPCR/Query/Builder/OperandDynamicFactory.php
```diff
@@ -59,9 +59,11 @@ public function fullTextSearchScore($alias)
      * $qb->where()
      *   ->gt()
      *     ->length('alias_1.prop_1')
-     *     ->literal(50);
+     *     ->literal(50)
+     *   ->end()
+     * ->end();
      *
-     * $qb->orderBy()->asc()->fullTextSearchScore('sel_1');
+     * $qb->orderBy()->asc()->fullTextSearchScore('sel_1')->end();
```
dantleech: again, `end` is only necessary when chaining more stuff

dantleech:
looks good, I would normally advise against adding the `end()` when it is not needed, as not having it simplifies things, but then not sure if that helps or hinders people (i.e. having to explain why it is not necessary there, but necessary when you have more stuff in the same line)

lib/Doctrine/ODM/PHPCR/Query/Builder/ConstraintFactory.php
dbu: i know, but we have them in some points but not others. i thought having them everywhere makes it easier to copy-paste fragments. do we have a doc section explaining when end() is needed? and does absolutely everything have end() or do we need to be more explicit with that?

lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterPhpcr.php
```diff
@@ -513,13 +515,16 @@ protected function doWalkConstraintComposite(AbstractNode $node, $method)
...
         $children = $node->getChildren();
 
-        if (count($children) == 1) {
-            $op = $this->dispatch(current($children));
-            return $op;
+        if (0 === count($children)) {
+            throw new \InvalidArgumentException('Composite must have at least one constraint');
+        }
+        if (1 === count($children)) {
+            return $this->dispatch(current($children));
```
dbu: good idea. also simplified the case of exactly one child a bit.

dbu:
is it ok if we merge this as is, and at some point somebody writes an introduction explaining the `->end()` bits? we need that either way.

my reason to always specify end() here is that people will use this doc to copy-paste fragments of their queries.

dantleech:
Yeah thats fine. We should [regenerate the generated documentation](https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DumpQueryBuilderReferenceCommand.php) after this is merged.

dbu:
good point. done in https://github.com/doctrine/phpcr-odm-documentation/commit/a5a56d8db7222fd002b0767f072541d5a01bf292

PR #620, opened 2015-03-25
Preserve order of operations

dantleech:
This PR is aiming to preserve the order of operations, it tries to do this whilst preserving the current behavior as much as possible.
- Everytime an INSERT, MOVE or DELETE is made, we add an Operation to the TreeOperationQueue
- No such action is taken for REORDER or UPDATE

When flush is called:
- The `execute*` methods are called with batches of operations, e.g. INSERT, INSERT, INSERT would be a single batch, however INSERT, INSERT, MOVE, INSERT would be three batches, so that the move is executed after the first 2 inserts.


lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -813,7 +800,7 @@ private function doScheduleInsert($document, &$visited, $overrideIdGenerator = n
                 // TODO: Change Tracking Deferred Explicit
                 break;
             case self::STATE_REMOVED:
-                unset($this->scheduledRemovals[$oid]);
+                $this->treeOpQueue->unqueue(TreeOperation::OP_REMOVE, $oid);
```
dantleech: Not 100% sure about the validity of this. If the Document was previously removed, and then re-inserted we remove all REMOVE operations from the stack. Not sure if that could conflict somehow with MOVE or REMOVE.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -927,16 +914,22 @@ public function scheduleMove($document, $targetPath)
 
         switch ($state) {
             case self::STATE_NEW:
-                unset($this->scheduledInserts[$oid]);
                 break;
             case self::STATE_REMOVED:
-                unset($this->scheduledRemovals[$oid]);
+                $this->treeOpQueue->unqueue(TreeOperation::OP_REMOVE, $oid);
```
dantleech: Same question as above

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -2318,7 +2324,9 @@ private function executeInserts($documents)
                     && !$class->isNullable($fieldName)
                     && !$this->isAutocreatedProperty($class, $fieldName)
                 ) {
-                    throw new PHPCRException(sprintf('Field "%s" of class "%s" is not nullable', $fieldName, $class->name));
+                    if (!$this->treeOpQueue->isQueued(TreeOperation::OP_REMOVE, $oid)) {
+                        throw new PHPCRException(sprintf('Field "%s" of class "%s" is not nullable', $fieldName, $class->name));
```
dantleech: We allow null fields on operations that are queued for removal because the changeset has been computed and unsert the fields, however we still need the node to be created in the timeline (it will be later removed).

dantleech:
@dbu any idea why this is not working on Jackrabbit? Something on our end?

dantleech:
````
--0935d494bb0a33542b1ed402934cca2a
--0935d494bb0a33542b1ed402934cca2a
--0935d494bb0a33542b1ed402934cca2a
Content-Disposition: form-data; name=":diff"
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

-/functional/user2 : buser : /functional/user1/subuserstructured","jcr:mixinTypes":["phpcr:managed"],"phpcr:class":"Doctrine\\Tests\\ODM\\PHPCR\\Functional\\User2","username":"test",}
--0935d494bb0a33542b1ed402934cca2a--
````

Resulting in:

````
PHPCR\PathNotFoundException: HTTP 409: /functional/user1/subuser
````

If I catch the `PathNotFoundException` the specific INSERT, MOVE, DELETE test works as expected.

File: lib/Doctrine/ODM/PHPCR/UnitOfWork.php (hunk 1 above)
Comment (dbu) [inline lib/Doctrine/ODM/PHPCR/UnitOfWork.php]: if you want to record operations, i think you would need to queue another add operation instead. i am sure it would be possible to find some hairbrained scenario where this makes a difference. and probably some very real usecase with listeners doing many things could actually result in the same scenario and actually make sense...

File: lib/Doctrine/ODM/PHPCR/UnitOfWork.php [hunk 4]
```diff
@@ -927,16 +914,22 @@ public function scheduleMove($document, $targetPath)
 
         switch ($state) {
             case self::STATE_NEW:
-                unset($this->scheduledInserts[$oid]);
```
dbu: this is a behaviour change i think, if we previously auto-persisted documents upon move operations.

dbu:
interesting! about the jackrabbit issue: can you try to do that in a phpcr-api-test (CombinedOperationsTest i guess) to see if we have a bug there?

i think there is a philosophical question we need to decide. as i understood, doctrine calculates the optimal path to transform an object model into database operations, and does not record changes procedurally. but if we don't manage to calculate this transition, maybe explicit ordered operations make more sense?

@Ocramius @lsmith77 what do you think about this?

PR #622, opened 2015-03-26
Allow sub localizations

dantleech:
It seems that locales such as `de_at` are not included when `->getLocalesFor` is called.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
+++ b/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
@@ -175,7 +175,7 @@ class AttributeTranslationStrategy extends AbstractTranslationStrategy
         $locales = array();
         foreach ($node->getProperties($this->prefix . ':*') as $prop) {
             $matches = null;
-            if (preg_match('/' . $this->prefix . ':([a-zA-Z1-9_]+)-[^-]*/', $prop->getName(), $matches)) {
+            if (preg_match('/' . $this->prefix . ':([a-zA-Z1-9_]+)-/', $prop->getName(), $matches)) {
                 if (is_array($matches) && count($matches) > 1 && !in_array($matches[1], $locales)) {
                     $locales[] = $matches[1];
                 }
--- a/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
+++ b/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
@@ -175,7 +175,7 @@ class AttributeTranslationStrategy extends AbstractTranslationStrategy
         $locales = array();
         foreach ($node->getProperties("*{$this->prefix}*") as $prop) {
             $matches = null;
-            if (preg_match('/' . $this->prefix . ':(..)-[^-]*/', $prop->getName(), $matches)) {
+            if (preg_match('/' . $this->prefix . ':([a-z]{2}_?[a-z]{0,2})-[^-]*/', $prop->getName(), $matches)) {
                 if (is_array($matches) && count($matches) > 1 && !in_array($matches[1], $locales)) {
                     $locales[] = $matches[1];
                 }
--- a/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
+++ b/lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
@@ -173,9 +173,9 @@ class AttributeTranslationStrategy extends AbstractTranslationStrategy
     public function getLocalesFor($document, NodeInterface $node, ClassMetadata $metadata)
     {
         $locales = array();
-        foreach ($node->getProperties("*{$this->prefix}*") as $prop) {
+        foreach ($node->getProperties($this->prefix . ':*') as $prop) {
             $matches = null;
-            if (preg_match('/' . $this->prefix . ':([a-z]{2}_?[a-z]{0,2})-[^-]*/', $prop->getName(), $matches)) {
+            if (preg_match('/' . $this->prefix . ':([a-zA-Z1-9_]+)-[^-]*/', $prop->getName(), $matches)) {
                 if (is_array($matches) && count($matches) > 1 && !in_array($matches[1], $locales)) {
                     $locales[] = $matches[1];
                 }
--- a/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/AttributeTranslationStrategyTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/AttributeTranslationStrategyTest.php
@@ -118,6 +118,74 @@ class AttributeTranslationStrategyTest extends \Doctrine\Tests\ODM\PHPCR\PHPCRFu
         $this->assertEquals(array(), $doc->getSettings());
     }
 
+    public function testSubRegionSaveTranslation()
+    {
+        // First save some translations
+        $data = array();
+        $data['topic'] = 'Some interesting subject';
+        $data['text'] ='Lorem ipsum...';
+
+        $node = $this->getTestNode();
+
+        $strategy = new AttributeTranslationStrategy($this->dm);
+        $strategy->saveTranslation($data, $node, $this->metadata, 'en');
+
+        // Save translation in another language
+
+        $data['topic'] = 'Some interesting american subject';
+
+        $strategy->saveTranslation($data, $node, $this->metadata, 'en_US');
+        $this->dm->flush();
+
+        // Then test we have what we expect in the content repository
+        $node = $this->session->getNode('/' . $this->testNodeName);
+
+        $this->assertTrue($node->hasProperty(self::propertyNameForLocale('en', 'topic')));
+        $this->assertTrue($node->hasProperty(self::propertyNameForLocale('en_US', 'topic')));
+        $this->assertTrue($node->hasProperty(self::propertyNameForLocale('en', 'text')));
+        $this->assertTrue($node->hasProperty(self::propertyNameForLocale('en_US', 'text')));
+        $this->assertFalse($node->hasProperty(self::propertyNameForLocale('en', 'author')));
+        $this->assertFalse($node->hasProperty(self::propertyNameForLocale('en_US', 'author')));
+
+        $this->assertEquals('Some interesting subject', $node->getPropertyValue(self::propertyNameForLocale('en', 'topic')));
+        $this->assertEquals('Some interesting american subject', $node->getPropertyValue(self::propertyNameForLocale('en_US', 'topic')));
+        $this->assertEquals('Lorem ipsum...', $node->getPropertyValue(self::propertyNameForLocale('en', 'text')));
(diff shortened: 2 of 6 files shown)


dantleech:
Would this PR be correct? If so I will add tests.

lsmith77:
seems sensible to me

lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
```diff
@@ -176,7 +176,7 @@ public function getLocalesFor($document, NodeInterface $node, ClassMetadata $met
         $locales = array();
         foreach ($node->getProperties("*{$this->prefix}*") as $prop) {
             $matches = null;
-            if (preg_match('/' . $this->prefix . ':(..)-[^-]*/', $prop->getName(), $matches)) {
+            if (preg_match('/' . $this->prefix . ':([a-z]{2}_?[a-z]{0,2})-[^-]*/', $prop->getName(), $matches)) {
```
dbu: there might be a BC break in this, as the previous expressoin was way more open. i would prefer to keep that the same, so probably `((..)(_(a-zA-Z){2})?)-[^-]*/` - please also add some tests about this. we should get the full locale in the match i think. 

btw, why do we forbid further `-` in the property name? we should probably just stop matching after the `locale-` thing, we don't care what comes. as we don't have a `$` at the end, it does not matter anyways. so `((..)(_(a-zA-Z){2})?)-`

also, i think we should allow upper case letters for the sub locale, often those are written as de_AT

dantleech: Hmm I understand that as accepting one of `xx`, `xx_xx` or `xx_` (unverified), I think yours does the same no?

For the upper case -- it would be better to normalize to lowercase upon creation I think.

dantleech: But yeah, it is untested, so whatever it should only match `xx_xx` and `xx` finally.

dbu: to be BC, we need to accept `<any><any>` or `<any><any>_xx`. look at the parenthesis, the question mark includes the _. as i want to accept any character in the first group, i made the sublocalization one optional instead. not sure about making all lowercase, but found no clear indication - but here for example, the country is in uppercase: http://php.net/manual/de/intl.examples.basic.php

dantleech: Hm, so are we saying that we want to preserve the users choice of captalization? I would prefer to normalize it to something rather than have `de_at` and `de_AT` coexisting (would make searches potentially unreliable f.e.).

I would say better to normalize it to lowercase and then the user can normalize it another way if they want. For BC I guess we would have to normalize here and when we create the property.

dbu: ok, agree that we should be consistent. in terms of BC, there can be any two characters or two characters followed by _xx which is only a-z lowercase. i would keep allowing non-letter locales in case people did that for some reason. it should not hurt imho.

dbu: dan, can you upate the expression to be more open to avoid potential BC breaks? apart from that i think we could merge it.

dbu:
ping @dantleech ;-)

lsmith77:
i added some testts

lsmith77:
actually not sure if we should allow any two chars. we could sell this as a bug fix imho :)

tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/AttributeTranslationStrategyTest.php
```diff
@@ -118,6 +118,74 @@ public function testLoadTranslation()
...
+        $data['topic'] = 'Some interesting american subject';
+
+        $strategy->saveTranslation($data, $node, $this->metadata, 'en_US');
+        $this->dm->flush();
+
+        // Then test we have what we expect in the content repository
+        $node = $this->session->getNode('/' . $this->testNodeName);
+
+        $this->assertTrue($node->hasProperty(self::propertyNameForLocale('en', 'topic')));
+        $this->assertTrue($node->hasProperty(self::propertyNameForLocale('en_US', 'topic')));
```
dbu: do you propose to go for uppercase countries? i don't know which is better, but we should be strict about it. i think dan's expression only allows lower case. either way i would add a test case for that.

btw, is it granted that there can only be two letter language and country names, or are there longer codes we need to respect? i.e. `zh-Hant` for chinese traditional...

dbu: actually it might make more sense to simply initialize the translation strategy with a whitelist of allowed (arbitrary) locales.

lsmith77: I guess we could also just use http://symfony.com/doc/current/reference/constraints/Locale.html

dantleech: I think you are right in that `fo_ba` is not enough: http://en.wikipedia.org/wiki/IETF_language_tag

Also everything can be considered as being lowercase I think: http://tools.ietf.org/html/rfc5646#section-2.1

I am not sure if we should add a whitelist -- could we reuse the existing language fallback mapping?

dbu: imho the most sensible thing to validate would be to check whether the locales match a configured set of expected locales. inside phpcr-odm they have no semantical meaning anyways. but they must match with fallback order and preferred locale. the fallback mapping indeed could serve as a whitelist. a language that does not occur in the list (you can add it with an empty fallback order if needed) is not known to the system...

lsmith77: not sure .. relying only on the configuration might lead to problems when you have multiple apps with different fallback rules (remember these can also be dynamic, so they can even change in a single application) talk to the same repository. it feels iffy to me ..

dbu: okay, i see that point. adding the symfony validator as a dependency for this seems quite big however. then again in a symfony project its already present so maybe not that bad...

lsmith77: it also has a hard dependency on the translation system but still .. I think its maybe better than trying to do our own.

dbu: should we make it an optional dependency but check to fail early when multilanguage is configured to say you need it? when not having multilang at all, its unneeded.

lsmith77: +1

dantleech:
Updated the regex to accept *any* characters from `[0-9a-zA-Z_]+`. I think validation should be done upon persist in anycase. The main thing is that the locale starts after the prefix and end before `-` I think.

tests/Doctrine/Tests/ODM/PHPCR/Translation/AttributeTranslationStrategyTest.php
```diff
@@ -8,20 +8,61 @@
...
+    }
+
+    public function testGetLocalesFor()
+    {
+        $classMetadata = $this->prophesize('Doctrine\ODM\PHPCR\Mapping\ClassMetadata');
+        $document = new \stdClass;
+        $localizedPropNames = array(
+            'test:de-prop1' => 'de',
+            'test:de_at-prop2' => 'de_at',
+            'test:en_Hans_CN_nedis_rozaj_x_prv1_prv2-prop3' => 'en_Hans_CN_nedis_rozaj_x_prv1_prv2',
```
dantleech: This example from http://php.net/manual/en/locale.composelocale.php

lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
```diff
@@ -173,9 +173,9 @@ public function removeAllTranslations($document, NodeInterface $node, ClassMetad
     public function getLocalesFor($document, NodeInterface $node, ClassMetadata $metadata)
     {
         $locales = array();
-        foreach ($node->getProperties("*{$this->prefix}*") as $prop) {
+        foreach ($node->getProperties($this->prefix . ':*') as $prop) {
```
dantleech: Note that I changed the pattern for `getProperties`. It was previously allowing text before the prefix and not requiring the prefix delimited `:`.  Is that correct?

dbu: sounds sane, yes. no idea why the old thing is like it was.

dbu:
the LOW version tests fail, but its the same on master.

i am +1 to merge this one, looks good to me.

lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
```diff
@@ -173,9 +173,9 @@ public function removeAllTranslations($document, NodeInterface $node, ClassMetad
...
             $matches = null;
-            if (preg_match('/' . $this->prefix . ':(..)-[^-]*/', $prop->getName(), $matches)) {
+            if (preg_match('/' . $this->prefix . ':([a-zA-Z1-9_]+)-[^-]*/', $prop->getName(), $matches)) {
```
dbu: is this correct? the more i look at this the more i get confused. what is the point of the `[^-]*` part? matching anything after the minus that is not a minus, but its optional and not even used afterwards... i think the original code was wrong here.

the new regex however will only match letters and numbers, and only if they are followed by _. the _ is optional. actually we should just check for something that comes after the prefix. and we have a naming issue in there - the thing called prefix can only be a namespace alias, not any prefix, because the `:` is hardcoded all over the place.

dbu: ach. howto regex. actually this does what it should. only the `[^-]` part is useless. will remove it.

tests/Doctrine/Tests/ODM/PHPCR/Translation/AttributeTranslationStrategyTest.php
```diff
@@ -8,20 +8,61 @@
...
         $this->assertEquals('test:fr-field', $name);
     }
+
+    public function testSubRegion()
+    {
+        $name = $this->method->invokeArgs($this->strategy, array('en_GB', 'field'));
+        $this->assertEquals('test:en_GB-field', $name);
+    }
+
+    public function testGetLocalesFor()
```
lsmith77: lets skip this test on PHP 5.3.3 since it seems to break there

PR #625, opened 2015-04-04
should depend on `doctrine/common ~2.4` instead fixed #624

### LINKED ISSUE #624: Dependency issue with doctrine/orm
Hi,

Doctrine ORM 2.5.0 has just been released, and now it explicitly requires `doctrine/common >= 2.5.x` while `doctrine/phpcr-odm` is explicitly requiring `doctrine/common ~2.4.0`. Is there a recommended way to resolve this conflict? Before I can use aliases but now it seems impossible.

I would avoid reverting to Doctrine ORM 2.4.0 at all cost because my current app depends on some of 2.5's features.

Thanks.

kminh:
fixed #624

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "minimum-stability": "dev",
     "require": {
         "php": ">=5.3.3",
-        "doctrine/common": "~2.4.0",
+        "doctrine/common": "~2.4",
         "phpcr/phpcr": "~2.1.1",
         "phpcr/phpcr-implementation": "~2.1.0",
         "phpcr/phpcr-utils": ">=1.1.0,<1.3.x-dev"


dbu:
looks like we get hit by the symfony deprecations now. @WouterJ do you have a suggestion how we should handle this? either we can copy https://github.com/symfony-cmf/Testing/blob/master/src/ErrorHandler/DeprecationErrorHandler.php to this codebase, or start dev-depending on symfony-cmf/testing. i tend towards copying the handler here. wdyt?

WouterJ:
@dbu I'm going to remove the error handler from the testing component soon. You should use [`symfony/phpunit-bridge`](http://github.com/symfony/phpunit-bridge) now. See https://github.com/sonata-project/SonataAdminBundle/commit/982a06f303bffc103ccafe83e33e386df9c02c5d for an example

dbu:
thanks @WouterJ . continued this in #626

PR #626, opened 2015-04-05
Fix deps

dbu:
Correct dependency on doctrine commons and use symfony phpunit handler to avoid failures due to deprecations

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -7,6 +7,14 @@ if (file_exists($file)) {
     throw new RuntimeException('Install dependencies to run test suite.');
 }
 
+$files = array_filter(array(
+    __DIR__.'/../vendor/symfony/symfony/src/Symfony/Bridge/PhpUnit/bootstrap.php',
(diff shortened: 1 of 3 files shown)


tests/bootstrap.php
```diff
@@ -7,6 +7,14 @@
     throw new RuntimeException('Install dependencies to run test suite.');
 }
 
+$files = array_filter(array(
+    __DIR__.'/../vendor/symfony/symfony/src/Symfony/Bridge/PhpUnit/bootstrap.php',
+    __DIR__.'/../vendor/symfony/phpunit-bridge/bootstrap.php',
+), 'file_exists');
+if ($files) {
+    require_once current($files);
+}
```
WouterJ: as this library will never install the `symfony/symfony` package during tests, we know that the bootstrap file is available in ``../vendor/symfony/phpunit-bridge/bootstrap.php``.

dbu: indeed. then again it should not hurt.

PR #629, opened 2015-04-10
Verify that uuid is only mapped on referenceable documents

dbu:
jackalope started to verify that the uuid is not set on a non-referenceable node, [leading to confusion](https://github.com/jackalope/jackalope/issues/281) when people map the uuid but do not make the document referenceable.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
@@ -441,6 +441,21 @@ class ClassMetadata implements ClassMetadataInterface
     }
 
     /**
+     * Validate whether this class needs to be referenceable.
+     *
+     * The document needs to be either referenceable or full versionable.
+     * Simple versioning does not imply referenceable.
+     *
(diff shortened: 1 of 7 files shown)


lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -441,6 +441,18 @@ public function validateIdentifier()
     }
 
     /**
+     * Validate whether this class needs to be referenceable.
+     *
+     * @throws MappingException if there is an invalid reference mapping
+     */
+    public function validateReferenceable()
+    {
+        if ($this->uuidFieldName && !$this->referenceable && !'full' === $this->versionable) {
```
dbu: simple versionable does not imply referenceable. not sure if this check is too restrictive as custom mixins could extend referenceable as well - but worst case you just need to explicitly set referenceable to true in that case.

lsmith77: I think we normally use ``!==``

dbu:
simplified the comparison statement.

PR #630, opened 2015-04-17
ignore null elements in reference / referrers collection when cascading ...

uwej711:
...remove

Background: I have some nested form collections which result in an entry in the ReferenceManyCollection with value null which I can't actually get rid off in another place.

Without this if, the code fails later when trying to get the spl_object_hash with a null value.

This corresponds to a similar if in cascadeScheduledInsert


lsmith77:
makes sense ..
the build issues are related to some issue we seem to have introduced in DoctrinePHPCRBundle and that we are also seeing in other repos.

dbu:
thanks a lot! the build error has to do with the version of doctrine dbal that gets installed, not the phpcr bundle - the bundle is not used by phpcr-odm.

PR #632, opened 2015-05-01
remove force option again

lsmith77:


tests/travis_doctrine_dbal.sh
```diff
@@ -19,5 +19,5 @@ fi
 mysql -e 'create database IF NOT EXISTS phpcr_odm_tests;' -u root
 
 cp ${SCRIPT_DIR}/../cli-config.doctrine_dbal.php.dist ${SCRIPT_DIR}/../cli-config.php
-${SCRIPT_DIR}/../bin/phpcrodm jackalope:init:dbal --force
+${SCRIPT_DIR}/../bin/phpcrodm jackalope:init:dbal
```
dbu: the problem with this is that the newer version of jackalope will need the force to actually do anything.

maybe we should just backport a --force option into old versions of jackalope (but doing nothing, that is executing even when --force is not specified)

lsmith77: https://github.com/jackalope/jackalope-doctrine-dbal/pull/263/files

PR #638, opened 2015-05-27
Introduced BuilderConverterBase to allow an extension point

dantleech:
This PR adds a base class for PHPCR query builder converters.


lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterBase.php
```diff
@@ -0,0 +1,576 @@
+<?php
+
```
dantleech: Missing header

lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterPhpcr.php
```diff
@@ -640,6 +455,8 @@ protected function walkConstraintNot(ConstraintNot $node)
         return $ret;
     }
 
+=======
+>>>>>>> Added base query builder converter
```
dantleech: Fix

lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
```diff
@@ -80,7 +81,7 @@ public function getQuery()
     /**
      * @param BuilderConverterPhpcr $converter
      */
-    public function setConverter(BuilderConverterPhpcr $converter)
+    public function setConverter(BuilderConverterBase $converter)
```
dbu: is there a chance that people extend the QueryBuilder? if they also overwrite this method, this would be a BC break. but i guess the risk for that is really small. current stable is 1.2 but master is 1.3 so that is ok imho

lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
```diff
@@ -7,6 +7,7 @@
 use Doctrine\ODM\PHPCR\Query\Query;
 use Doctrine\ODM\PHPCR\Query\Builder\AbstractNode as QBConstants;
 use Doctrine\ODM\PHPCR\Exception\RuntimeException;
+use Doctrine\ODM\PHPCR\Query\Builder\BuilderConverterBase;
```
dbu: i think this is not needed, its in the same namespace

dbu:
looks correct to me. can you explain the use case for this?

dantleech:
I built a different document manager implementation in Sulu, but it can reuse most of the query builder logic. I guess in terms of "legitimate" use cases, it provides a starting point for people who want to extend the grammer of the query builder.

lsmith77:
@dantleech how did you choose which methods to implement in the abstract class?

dantleech:
@lsmith77 all the methods in the abstract class are (or are supposed to be) agnostic to the PHPCR-ODM.  I did it by requirement, we might be able to fill in some of the abstract methods with some base functionality, but I guess its not too important.

If you agree with this change, its probably worth waiting until I have finished the our integration before merging.

lsmith77:
yeah .. I guess we can wait until your use case is ready and then decide, ok?

dbu:
+1, so we leave this open until you confirm its good to merge from your POV

lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
```diff
@@ -80,7 +80,7 @@ public function getQuery()
     /**
      * @param BuilderConverterPhpcr $converter
      */
-    public function setConverter(BuilderConverterPhpcr $converter)
+    public function setConverter(BuilderConverterInterface $converter)
```
dantleech: Introduced an interface, seems more bullet proof.

dantleech:
@dbu @lsmith77 this seems good from our side. I added a `BuilderConverterInterface` instead of type hinting on `BuilderConverterBase`.

lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\ODM\PHPCR\Query\Builder;
+
+interface BuilderConverterInterface
```
dbu: can you please add a phpdoc what exactly a BuilderConverter is? can we find a better name? like maybe QueryBuilderTransformerInterface?

lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
dbu: was going to say this is a BC break but actually not, as BuilderConverterPhpcr was more restricted. that class implements the interface so people extending BuilderConverterPhpcr would still be able to call this method with their extending class.

dbu:
the lower version of tests is failing. not sure if we need to just increase the minimal requirements or if we introduce a nasty BC thing here. can you please have a look?

lib/Doctrine/ODM/PHPCR/Query/Builder/BuilderConverterInterface.php
dantleech: Hmm, I think `Converter` is synonymous with `Transformer`, and the existing implementation is `Converter`. Not sure about prefixing `Query`. Prefixing `Builder` is also questionably, as we are just repeating contextual information that is already in the namespace. Also renaming the interface would break the alphabetic grouping unless we also rename the `BuilderConverterPhpcr` class.

dbu: okay. i would prefer `ConverterInterface`, `BuilderConverterInterface` is rather arbitrary.

dantleech: Then, by extension, we should also rename the `BuilderConverterPhpcr` class to `ConverterPhpcr`.

dbu: agreed. are these classes ever used outside our own things? if so we should keep an empty BC class, otherwise lets just drop it.

i suggest we go to the next minor version in composer.json and add a changelog that the class was renamed. i don't think its supposed to be used outside the library.

dantleech:
ok, have removed the `Builder` previx.

lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterBase.php
```diff
@@ -1,93 +1,105 @@
...
 use Doctrine\ODM\PHPCR\Exception\RuntimeException;
 use Doctrine\ODM\PHPCR\Mapping\ClassMetadata;
 use Doctrine\ODM\PHPCR\Translation\TranslationStrategy\TranslationStrategyInterface;
+use PHPCR\Query\QOM\EquiJoinConditionInterface;
+use PHPCR\Query\QOM\QueryObjectModelFactoryInterface;
 use Doctrine\ODM\PHPCR\Mapping\ClassMetadataFactory;
+use PHPCR\Query\QOM\QueryObjectModelConstantsInterface as QOMConstants;
 use Doctrine\ODM\PHPCR\Query\Query;
-use Doctrine\ODM\PHPCR\DocumentManagerInterface;
+use Doctrine\ODM\PHPCR\DocumentManager;
```
dbu: why change this back to not the interface?

dbu: i think this was a recent refactoring. maybe a rebase conflict resolution that went wrong?

dbu:
cool! can you squash the commits and look at DocumentManager vs DocumentManagerInterface?

dantleech:
Hmm. The `DocumentManagerinterface` was not used, but would indicate that I didn't use the latest version of the class .. maybe worth doing the split again.

Removed all of the unused use statements using `phpcsfixer`.

dbu:
maybe you don't need to use either anymore. but lukas changed things to use DocumentManagerInterface - if we typehint anything about DocumentManager in these classes, we need to investigate. if its unused anyways, it might just have been a global search and replace that is not relevant here.

dantleech:
Yeah, the query builder code only refers to the `DocumentManagerInterface`. I have done a quick sanity check and can't see that we have lost anything else.

dbu:
:+1: 
can you squash the commits? then i think we are good to merge.

dantleech:
Squashed

PR #643, opened 2015-06-19
Fix PHP 7 compatibility

### LINKED ISSUE #641: PHP 7.0 incompatibility
```
PHP Fatal error:  Cannot use 'String' as class name as it is reserved in /home/travis/build/liip-forks/symfony-standard/vendor/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/Mapping/Annotations/String.php on line 28
```

WouterJ:
Fixes #641

This is as decided on IRC this morning by @dantleech, @lsmith77 and me.

Open Questions
---

 * Should we rename `TranslatedProperty` to `Field`, or keep the current way?
 * Should we trigger deprecation notices?


WouterJ:
@dantleech to answer you question on IRC (*WouterJ: isn't it missing the annotation driver changes?*): There is no need for annotation driver changes, as it bases on `Annotations\Property`. Effectively, a class `Field` with property `$type` set to `string` is the same as a class `String` with property `$type` set to `string` by default.

WouterJ:
Tests are green, even for PHP 7. So this PR can be merged.

I propose to leave PHP 7 in allowed_failures for some months, as it's still a nightly version, so errors still can be introduced in newer PHP 7 builds.

tests/bootstrap.php
```diff
@@ -17,7 +17,7 @@
 
 use Doctrine\Common\Annotations\AnnotationRegistry;
 AnnotationRegistry::registerLoader(array($autoload, 'loadClass'));
-AnnotationRegistry::registerFile(__DIR__.'/../lib/Doctrine/ODM/PHPCR/Mapping/Annotations/DoctrineAnnotations.php');
+//AnnotationRegistry::registerFile(__DIR__.'/../lib/Doctrine/ODM/PHPCR/Mapping/Annotations/DoctrineAnnotations.php');
```
lsmith77: why not remove this?

lsmith77:
I guess in theory we could now also import ``Field`` so that we can just write ``@Field`` but well .. not sure if we should do this in the existing tests ...

tests/bootstrap.php
dantleech: +1

dantleech:
I would keep the `PHPCRODM` prefix for consistency, in some of the tests other annotations such as `@Node` are used.

WouterJ:
In fact, each document in the tests contains `@PHPCRODM\Id`.

tests/bootstrap.php
dbu: see also the discussion on the cmf mailinglist. do we need the line for `loadClass` even?

dbu:
having a prefix seems best practice to me, as it will better report
typos. but afaik the convention of the doc is to use `@PHPCR\...` so we
should go with that and be consistent. does not need to be in this PR
however.

tests/bootstrap.php
WouterJ: Removed the line. I'm not sure about removing registerLoader. I know Symfony registers `class_exists` as an annotation loader, so it kind of  seems like you need to configure at least one loader to make things work. Let's do it in another PR...

lsmith77:
good to merge imho

lsmith77:
BTW I think we should then adjust all repos to use ``Field`` adding a requirement for phpcr odm 1.3

WouterJ:
I agree, this would mean that this "feature" (bug fix?) will be released in 1.3.1?

lsmith77:
we have not released 1.3.0

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Binary.php
```diff
@@ -21,9 +21,13 @@
 
 use Doctrine\Common\Annotations\Annotation;
 
+@trigger_error('The @Binary annotation is deprecated, use @Field(type="binary") instead.', E_USER_DEPRECATED);
```
dbu: do we need to trigger_error? this was a huge mess with symfony, when suddenly tests start to fail because of deprecations. if we do we should at least explain in the changelog how to use the symfony deprecation handle to have succeeding tests even with those deprecations.

otherwise i vote to not trigger deprecation errors for now.

WouterJ: Symfony changed its error strategy in 2.7.1, which is what I'm using here too. As the error is silenced, it isn't visible when using a normal PHP installation. It'll only be visible when using a custom error handler (like the one provided with the Symfony PHPunit bridge or xDebug).

This means that these errors are now opt-in instead of the previous "opt-out"-ish situation of 2.7-dev. That's why I've added it here, but I'm open to comment this line if you don't think it's the correct moment to trigger errors

dbu: what if they are still on symfony 2.3 for example, but update phpcr-odm?

and i don't know if anybody is using phpcr-odm outside of symfony - but if we have an easy answer for them i am happy to find out :-)

WouterJ: It's not about the Symfony version you use, it's about the addition of `@` in front of `trigger_error` (whereas Symfony 2.7-dev didn't have this `@` prefix). This means the error is silenced, unless you have a specific error handler for this type of errors.

Anyway, I'll comment this out (I want to have this PR merged asap).

WouterJ:
Cool! :smile_cat: 

Btw, this only applies to annotations. As all CMF bundles use XML mapping, we don't have to update much repos (just SE and Sandbox)

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Float.php
```diff
@@ -21,10 +21,15 @@
...
 
+//@trigger_error('The @Float annotation is deprecated, use @Field(type="double") instead.', E_USER_DEPRECATED);
+
 /**
  * Convenience alias for Double.
+ *
  * @Annotation
  * @Target("PROPERTY")
+ *
+ * @deprecated Use @Field(type="double") instead.
```
lsmith77: is that a typo?

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Int.php
```diff
@@ -21,10 +21,15 @@
...
 
+//@trigger_error('The @Int annotation is deprecated, use @Field(type="long") instead.', E_USER_DEPRECATED);
+
 /**
  * Convenience alias for Long.
+ *
  * @Annotation
  * @Target("PROPERTY")
+ *
+ * @deprecated Use @Field(type="long") instead.
```
lsmith77: is that a typo?

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Float.php
WouterJ: No, as noted by the PHPdoc (line 27): "Convenience alias for Double"

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Int.php
WouterJ: "Convenience alias for Long" :)

lsmith77:
just realized .. we also need to update the docs ..
https://github.com/doctrine/phpcr-odm-documentation/issues/73

dantleech:
Added documentation for `@Fileld` in https://github.com/doctrine/phpcr-odm-documentation/pull/74

PR #644, opened 2015-06-25
Allow changeset computations of removed documents

### LINKED ISSUE #637: Flushing one removed document throws an exception
The following code

```php
$manager->remove($resource);
$manager->flush($resource);
```
throws an exception in the UnitOfWork class
```
Document has to be managed for single computation Symfony\Cmf\Bundle\MenuBundle\Doctrine\Phpcr\MenuNode@0000000075dd08b200000000252e59a6 (/be/menus/footer/press)
```

This is caused by the check in `Doctrine\ODM\PHPCR\UnitOfWork::computeSingleDocumentChangeSet()`

```php
public function computeSingleDocumentChangeSet($document)
{
    $state = $this->getDocumentState($document);
    if ($state !== self::STATE_MANAGED) {
        throw new InvalidArgumentException('Document has to be managed for single computation '.self::objToStr($document, $this->dm));
    }
    //...
}
```

Should this check be changed to keep STATE_REMOVED in mind?

I'm happy to provide a pull request, if this is the way to go. The current solution is to call flush without arguments.

coudenysj:
This fixes #637.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -1090,7 +1090,8 @@ class UnitOfWork
     public function computeSingleDocumentChangeSet($document)
     {
         $state = $this->getDocumentState($document);
-        if ($state !== self::STATE_MANAGED) {
+
+        if ($state !== self::STATE_MANAGED && $state !== self::STATE_REMOVED) {
             throw new InvalidArgumentException('Document has to be managed for single computation '.self::objToStr($document, $this->dm));
         }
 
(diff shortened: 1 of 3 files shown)


coudenysj:
@lsmith77 What can I do to let the dbal test pass? https://travis-ci.org/doctrine/phpcr-odm/jobs/68288848

lsmith77:
I think the issue is that we did some changes to the DoctrinePhpcrBundle which causes issues with travis that we only fixed in master. I will have a look.

lsmith77:
or if you have time .. look at the differences in the composer.json and .travis.yml files

coudenysj:
Will do

coudenysj:
Output from travis is _"ATTENTION: This operation should not be executed in a production environment. Please use "--force" to execute the command."_.

I think it has to do with the `InitDoctrineDbalCommand.php` file.

```diff
-        "jackalope/jackalope-doctrine-dbal": "~1.0",
+        "jackalope/jackalope-doctrine-dbal": "~1.1",
```

Maybe https://github.com/jackalope/jackalope-doctrine-dbal/commit/d55c96d8f3007f6f2c77448da982cdb6be8b994b should be applied to 1.0 too?

lsmith77:
can you add ``"jackalope/jackalope-doctrine-dbal": "~1.1"`` to the ``require-dev`` section of this PR?

coudenysj:
@lsmith77 Looks like fb77d60 did the trick (I forgot to commit is, locally I had to add it too to run the tests).

Issue #640, opened 2015-06-12
Mixins in xml mapping are not properly inherited


uwej711:
Example: class A defines mixin a, class B extends A and defines mixin b
Expected result: class B has mixins a and b, but you have to define mixin a again on class B to get the desired result.

Trying to add a failing test.

dbu:
i think that was kind of on purpose to allow preventing to inherit a mixin.
though semantically it makes little sense to have a child class not have a
mixin of the parent. we should maybe have an option to not inherit or
remove specific mixins and inherit by default. there is a BC problem
however. maybe we can do it as minor versiin, with a warning in changelog.
Am 12.06.2015 09:02 schrieb "Uwe Jäger" <notifications@github.com>:

dantleech:
What would PHPCR do in such a case? i.e.

````
[mytype] < nt:unstructured, somemixin

[myinheritedtype] < mytype
````

I think `myinheritedtype` would inherit the mixin.

uwej711:
I think the issue is that we don't "merge" the mixins from class A and B and just look at the mixins. If I remember right, class B will inherit  mixin a as long as class B does not define another mixin. I need to  investigate that bit and check whether it is an issue with the XmlDriver or with ClassMetadata.

dbu:
i think your analysis is correct uwe. and i think merging by default
would be right. its probably a problem with all mapping drivers.

dbu:
@uwej711 do you have time to look into this anytime soon? would be good if we could include this in the upcoming 1.3 version of phpcr-odm.

uwej711:
I had a look at this (and changed the implementation) but now I have a failing test:
```ClassMetadataFactoryTest::testLoadMetadataClassInheritanceChildCanOverride()``` which assumes that a child class can override the document definitions of the parent class.

So either we leave this as it is or change the failing tests (i.e. you can not override the parent class' mixins) or we add an option to the mixins definition to let the developer decide whether mixins should be merged or overridden.

dbu:
hm, could we do the options `add-mixins` and `replace-mixins` and
deprecate the use of `mixins`? i guess the other options are implicit
replaces, e.g. the primary type? i think adding mixins is more expected
and natural, but that would be a BC break. then again it should rarely
be a problem if an extending class has additional mixins so maybe just
keep `mixins` and change it to adding, and provide an additional
`replace-mixins` for when replacing is wanted.

uwej711:
[...]

PR #648, opened 2015-07-07
tests and changes for issue 640

uwej711:
See #640

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Changelog
 1.3.0
 -----
 
+* **2015-07-09**: Changed the semantics for mixins from overriding parent classes' mixins
+                  to adding mixins to already existing mixins. Introduced an alternative
+                  mapping "<replace-mixins>" for XML- and "replaceMixins" for YML- and Annotations-
+                  Mapping with the old semantics.
+                  If you depend on the old behaviour make sure to change your mixins mapping to
+                  replace-mixins or replaceMixins.
+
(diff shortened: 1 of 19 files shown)


lsmith77:
```Doctrine\Tests\ODM\PHPCR\Mapping\ClassMetadataFactoryTest::testLoadMetadataClassInheritanceChildCanOverride`` seems to be failing

I guess we will put this into 1.3 only as it can maybe lead to some surprises .. its the bug fix type than can feel like a BC break ... right?

uwej711:
Well it breaks BC but chances are only few people rely on the current behaviour.

dbu:
the discussion is in #640 - i agree with considering this a bugfix with potential to break somebody relying on the bug.

@uwej711 do you want to add the replace-mixins attribute here or in a separate pull request?

uwej711:
I would like to add it here ...

dbu:
:+1:

uwej711:
So I actually did this differently than I thought before, having mixins and replaceMixins in ClassMetadata would have made it necessary to change ClassMetadataFactory  and handling the replace logic there (which I now realize actually makes some sense). Maybe I will try that approach, too.

dbu:
i like this implementation, imo we can merge as is. do you think the other aproach would be much better?

uwej711:
The changes to this implementation I think about would move the check for mixins and replace-mixins from each driver to one implementation in ClassMetadata and the logic for overwriting the parent class' mixins should go to ClassMetadataFactory. These changes would reduce some redundancy and place the replace logic where the actual inheritance happens.

uwej711:
closed in favour of #650

PR #650, opened 2015-07-12
Mixin inheritance alternative

uwej711:
This is an alternative to #648 with less code duplication in the drivers and the replace or merge logic where it belongs. What I don't like with this solution is the replaceMixins attribute in ClassMetadata which is used only temporary.


dantleech:
Can you use both `replaceMixins` and `mixins` concurrently? If not I would vote for making `replaceMixins` a boolean and reusing the existing `mixins` key.

dbu:
indeed this is better. either way we have a bit of dual use of `mixins` as what is configured on the class itself and later it contains the merged mixins.

regarding dans proposal: i think that idea is good. maybe we could call the property `inheritMixins` and phpdoc that this is controlled by using `replace-mixins`. for the mapping, it might also make more sense to have a separate flag inherit-mixins. (e.g. in the edge case that i want to remove all mixins, i would need `replace-mixins=""` which feels strange). so it would always be the mixins mapping, and inherit-mixins defaults to true but can be false in which case only local mixins (if any) are used.

uwej711:
If we go for the boolean, I think it makes sense to have a boolean flag in the mapping as well. That would be 
<mixins inherit="true"> for XML but an extra flag for annotations and YML.

Cheers
Uwe

dbu:
sounds right, yes!

lsmith77:
ok so we will go with a single tag for the mixins with the attribute to prevent **automatic merging** of the parent mixins, ie. ``<mixins inherit="true">``.

this means if someone changes a parent to add mixins, they should clearly document this. ideally they should provide a migration to update existing documents anyway. if I then pull in the changes to this parent, I should see this in the changelog and then decide if I from now on what to manage the mixins myself or if I want to inherit automatically. seems like a sane workflow to me.

@uwej711 can you update one of your PRs accordingly?

dbu:
> can you update one of your PRs accordingly?

i prefer the approach in this PR, less code duplication and having the info whether to inherit or not in the metadata seems good to me.

dbu:
@uwej711 i don't want to rush you, but we hope to tag a RC1 like tomorrow. any chance you find a moment to wrap this up? otherwise i can have a go at it, but don't want to "steal" this one from you ;-)

uwej711:
OK, I will try to get this ready till tomorrow :-)

Cheers
Uwe

uwej711:
done - I won't be able to work on this tomorrow (i.e. today), so feel free to fix anything I missed before tagging RC1

lsmith77:
looks good to me.

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadataFactory.php
```diff
@@ -133,6 +133,10 @@ protected function doLoadMetadata($class, $parent, $rootEntityFound, array $nonS
             $this->getDriver()->loadMetadataForClass($class->getName(), $class);
         }
 
+        if ($parent && $class->getInheritMixins()) {
+            $class->setMixins(array_merge($parent->getMixins(), $class->getMixins()));
+        }
```
dbu: i think this should be done inside `addInheritedDocumentOptions`

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadataFactory.php
```diff
@@ -155,7 +159,6 @@ private function addInheritedDocumentOptions(ClassMetadata $subClass, ClassMetad
         $subClass->setVersioned($parentClass->versionable);
         $subClass->setReferenceable($parentClass->referenceable);
         $subClass->setNodeType($parentClass->getNodeType());
-        $subClass->setMixins($parentClass->getMixins());
```
dbu: @lsmith77 do you see the reason why this did not always let the mixins of the parent class win and prevent ever overwriting the mixins? same actually for the other options - we do not check if something is already set...

dbu:
thanks a lot! i can move the code to where setMixins was called before, and update the documentation.

but i am really confused reading addInheritedDocumentOptions - i dont see how it would not always overwrite whatever you set on the child document itself.

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadataFactory.php
lsmith77: yeah looks like it would just cause child mixins to be ignored ..

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadataFactory.php
uwej711: adding inherited mapping happens before the class' metadata is loaded, so this needs to happen here

Cheers
Uwe

uwej711:
This happens before the class metadata is loaded, therefore it works.

Mit freundlichen Grüßen
Uwe Jäger

dbu:
thanks uwe for the explanation. now i get it. i am adding a bit to the doc, squashing the commits and then will merge on the cli.

dbu:
merged manually as 0bf143026954cc9438dffb5ebea1c30bc10d0a7e

thanks a lot uwe!

PR #651, opened 2015-07-15
make Property an annotation, instead of adding new annotaiton Field

dbu:
see https://github.com/doctrine/phpcr-odm-documentation/pull/74

we need no BC for `@Field` as it was never released.

while doing this, i noticed that in yml mapping, the list of properties is called `fields` which would be an argument in favour of using `@Field` after all. but `@Property` seems more natural for the annotations for me.


lib/Doctrine/ODM/PHPCR/Mapping/Annotations/TranslatedProperty.php
```diff
@@ -18,13 +18,9 @@
...
-
-use Doctrine\Common\Annotations\Annotation;
-
 /**
- * @Annotation
- * @Target("PROPERTY")
+ * @deprecated Use Property instead
  */
-final class Field extends TranslatableProperty
+class TranslatedProperty extends Property
```
dbu: its just git thinking that this is a rename. i delete Field and rename TranslatedProperty to Property, but add this empty class for BC in case some custom code was typechecking TranslatedProperty for something.

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/TranslatableProperty.php
```diff
@@ -19,12 +19,8 @@
 
 namespace Doctrine\ODM\PHPCR\Mapping\Annotations;
 /**
- * Base class for all the translatable properties (i.e. every property but Uuid and Version)
+ * @deprecated Use Property instead
  */
 class TranslatableProperty extends Property
```
dbu: keeping for BC in case somebody extended it or has custom code that typechecks.

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Uuid.php
```diff
@@ -25,7 +25,7 @@
  * @Annotation
  * @Target("PROPERTY")
  */
-final class Uuid extends Property
+final class Uuid extends BaseProperty
```
dbu: if somebody was typechcking Property in custom code handling annotations, this is a BC break. do we need to worry?

lsmith77: i would say no

lsmith77:
Changelog also needs to be updated

lsmith77:
tests failing due to
```PHP Fatal error:  Class 'Doctrine\Tests\ODM\PHPCR\Mapping\Model\PropertyMappingObject' not found in /home/travis/build/doctrine/phpcr-odm/tests/Doctrine/Tests/ODM/PHPCR/Mapping/AbstractMappingDriverTest.php on line 22```

dbu:
ups. fixed, and tests where green. now updated the changelog.

@dantleech @WouterJ @lsmith77 ok with going like this? if so, i will update the doc as well.

lsmith77:
+1 from me

WouterJ:
I'm -1 on this change, but as it's 3 vs 1, I think I have no option...

It's inconsistent with yaml, mongodb odm and couchdb odm. Also, it's a BC break for people having a custom annotation extending from Property (it know suddently becomes translatable, while the function of Property was to have something that's not translatable).

dbu:
oh, so mongodb and couchdb use Field? that would indeed be a strong argument in favor of using Field here as well. 

loosk like: http://doctrine-couchdb.readthedocs.org/en/latest/reference/basic-mapping.html#field-mapping

so i am :-1: on my own PR :-/

lsmith77:
alright .. then i will reactivate my docs PR to remove ``@Property``

lsmith77:
merged https://github.com/doctrine/phpcr-odm-documentation/pull/75

PR #655, opened 2015-08-25
add a tool to convert translations

dbu:
fix https://github.com/symfony-cmf/CoreBundle/issues/164

doc PR: https://github.com/doctrine/phpcr-odm-documentation/pull/77

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "doctrine/data-fixtures": "~1.0,>=1.0.0",
         "phpcr/phpcr": "~2.1.1",
         "phpcr/phpcr-implementation": "~2.1.0",
-        "phpcr/phpcr-utils": "~1.2.7|~1.3",
+        "phpcr/phpcr-utils": "^1.2.8",
         "doctrine/instantiator": "~1.0.1",
         "symfony/console": "~2.3|~3.0"
     },
--- a/lib/Doctrine/ODM/PHPCR/Query/Query.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Query.php
@@ -25,7 +25,15 @@ class Query
     protected $firstResult;
     protected $maxResults;
     protected $documentClass;
+
+    /**
+     * @var QueryInterface
+     */
     protected $query;
+
+    /**
+     * @var DocumentManagerInterface
+     */
     protected $dm;
 
     public function __construct(QueryInterface $query, DocumentManagerInterface $dm, $primaryAlias = null)
@@ -197,6 +205,8 @@ class Query
      * Gets the list of results for the query.
      *
      * Alias for execute(null, $hydrationMode = HYDRATE_DOCUMENT).
+     *
+     * @param int $hydrationMode One of the hydration class constants
      *
      * @return array
      */
--- a/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
+++ b/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
@@ -52,7 +52,8 @@ class DocumentConvertTranslationCommand extends Command
 
             ->addArgument('classname', InputArgument::REQUIRED, 'Class that has changed translation information')
             ->addOption('previous-strategy', 'prev', InputOption::VALUE_OPTIONAL,
-                'Name of the previous translation strategy if there was one. Omit for converting from non-translated to translated'
+                'Name of the previous translation strategy if there was one. Omit for converting from non-translated to translated',
+                'none'
             )
             ->addOption('fields', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL,
                 'The fields to convert. If not specified, all fields configured as translated will be converted.',
--- a/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
+++ b/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
@@ -103,7 +103,7 @@ HERE
         $previous = $input->getOption('previous-strategy');
         $fields = $input->getOption('fields');
 
-        while ($converter->convert($class, $previous, $fields)) {
+        while ($converter->convert($class, $fields, $previous)) {
             $session->save();
             $output->write('.');
         }
--- a/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
+++ b/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
@@ -109,6 +109,8 @@ HERE
         }
 
         $output->writeln('done');
+
+        return 0;
     }
 
     /**
--- a/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
+++ b/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
@@ -59,6 +59,10 @@ class DocumentConvertTranslationCommand extends Command
                 'The fields to convert. If not specified, all fields configured as translated will be converted.',
                 array()
             )
+            ->addOption('locales', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL,
+                'Locales to copy previously untranslated fields into.',
+                array()
+            )
             ->addOption('force', null, InputOption::VALUE_NONE, 'Use to bypass the confirmation dialog')
             ->setHelp(<<<HERE
 The <info>doctrine:phpcr:docment:convert-translation</info> command migrates translations
@@ -82,6 +86,7 @@ HERE
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $class = $input->getArgument('classname');
+        $locales = $input->getOption('locales');
         $force = $input->getOption('force');
         if (!$force) {
             $force = $this->askConfirmation(
@@ -104,7 +109,18 @@ HERE
         $previous = $input->getOption('previous-strategy');
         $fields = $input->getOption('fields');
 
-        while ($converter->convert($class, $fields, $previous)) {
+        while ($converter->convert($class, $locales, $fields, $previous)) {
+            $notices = $converter->getLastNotices();
+            if (count($notices)) {
+                foreach ($notices as $path => $realClass) {
+                    $output->writeln(sprintf(
+                        'Document at %s is of class %s but requested to convert %s.',
+                        $path,
+                        $realClass,
+                        $class
+                    ));
+                }
+            }
             $session->save();
             $output->write('.');
         }
--- a/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
+++ b/lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
@@ -1,0 +1,147 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\ODM\PHPCR\Tools\Console\Command;
+
+use Doctrine\ODM\PHPCR\Tools\Helper\TranslationConverter;
+use PHPCR\SessionInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+/**
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+class DocumentConvertTranslationCommand extends Command
+{
+    private $translationConverter;
+
+    public function __construct($name = null, TranslationConverter $translationConverter = null)
+    {
+        parent::__construct($name);
+        $this->translationConverter = $translationConverter;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('doctrine:phpcr:document:convert-translation')
+            ->setDescription('Convert fields to translated or back to untranslated, and between different strategies after a refactoring.')
+
+            ->addArgument('classname', InputArgument::REQUIRED, 'Class that has changed translation information')
+            ->addOption('previous-strategy', 'prev', InputOption::VALUE_OPTIONAL,
+                'Name of the previous translation strategy if there was one. Omit for converting from non-translated to translated'
+            )
+            ->addOption('fields', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL,
+                'The fields to convert. If not specified, all fields configured as translated will be converted.',
+                array()
+            )
+            ->addOption('force', null, InputOption::VALUE_NONE, 'Use to bypass the confirmation dialog')
+            ->setHelp(<<<HERE
+The <info>doctrine:phpcr:docment:convert-translation</info> command migrates translations
+from a previous format to the current mapping.
+
+  <info>$ php ./app/console/phpcr doctrine:phpcr:document:convert-translation "Document\\ClassName"</info>
+
+<comment>When some fields already where translated, you need to specify which fields to convert.</comment>
+Failing to do that would erase all fields already translated previously.
+
+Note that when only some fields changed or when converting between translation strategies, you need
+to specify the previous strategy. When converting to untranslated, you additionally need to specify
+the fields that previously where translated.
+HERE
+            );
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $class = $input->getArgument('classname');
+        $force = $input->getOption('force');
+        if (!$force) {
+            $force = $this->askConfirmation(
+                $input,
+                $output,
+                sprintf('<question>Are you sure you want to migrate translations in %s Y/N ?</question>', $class),
+                false
+            );
+        }
+
+        if (!$force) {
+            $output->writeln('<error>Aborted</error>');
+
+            return 1;
+        }
+
+        /** @var $session SessionInterface */
+        $session = $this->getHelper('phpcr')->getSession();
+        $converter = $this->getTranslationConverter();
+        $previous = $input->getOption('previous-strategy');
+        $fields = $input->getOption('fields');
+
+        while ($converter->convert($class, $previous, $fields)) {
+            $session->save();
+            $output->write('.');
+        }
+
+        $output->writeln('done');
+    }
+
+    /**
+     * @return TranslationConverter
+     */
+    private function getTranslationConverter()
+    {
+        if (!$this->translationConverter) {
+            $this->translationConverter = new TranslationConverter(
+                $this->getHelper('phpcr')->getDocumentManager()
+            );
+        }
+
+        return $this->translationConverter;
+    }
+    /**
+     * Ask for confirmation with the question helper or the dialog helper for symfony < 2.5 compatibility.
+     *
+     * @param InputInterface  $input
+     * @param OutputInterface $output
+     * @param string          $question
+     * @param boolean         $default
+     *
+     * @return string
+     */
+    private function askConfirmation(InputInterface $input, OutputInterface $output, $question, $default = true)
+    {
+        if ($this->getHelperSet()->has('question')) {
+            $question = new ConfirmationQuestion($question, $default);
+
+            return $this->getHelper('question')->ask($input, $output, $question);
+        }
+
+        return $this->getHelper('dialog')->askConfirmation($output, $question, $default);
+    }
+}
--- a/lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
+++ b/lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
@@ -48,7 +48,7 @@ class TranslationConverter
     }
 
     /**
-     * Migrate content into the new translation form and remove the old properties.
+     * Migrate content into the new translation format and remove the old properties.
      *
      * This does not commit the changes to the repository. Call save on the
      * PHPCR session *after each batch*. Calling flush on the document manager
@@ -64,11 +64,20 @@ class TranslationConverter
      * To convert a single field from translated to non-translated, simply
      * specify that field.
      *
+     * If you convert an existing translation, you need to specify the name of
+     * the strategy that was previously used. The name is the one you would use
+     * for DocumentManagerInterface::getTranslationStrategy, so "attribute" or
+     * "child".
+     *
+     * The current strategy is read from the document metadata. If you need,
+     * you can however specify the strategy explicitly.
+     *
      * @param string       $class                FQN of the document class
-     * @param string|null  $previousStrategyName Name of previous strategy or null if field was not
-     *                                           previously translated
      * @param array        $fields               List of fields to convert. Required when making a
      *                                           field not translated anymore
+     * @param string|null  $previousStrategyName Name of previous strategy or null if field was not
+     *                                           previously translated
+     * @param string|null  $currentStrategyName  Name of current strategy or null to auto discover
      *
      * @return boolean true if there are more documents to convert and this method needs to be
      *                      called again.
@@ -77,12 +86,15 @@ class TranslationConverter
      */
     public function convert(
         $class,
+        array $fields = array(),
         $previousStrategyName = null,
-        array $fields = array()
+        $currentStrategyName = null
     ) {
         /** @var ClassMetadata $currentMeta */
         $currentMeta = $this->dm->getClassMetadata($class);
-        $currentStrategyName = $currentMeta->translator;
+        if (!$currentStrategyName) {
+            $currentStrategyName = $currentMeta->translator;
+        }
 
         // sanity check strategies
         if ($currentStrategyName === $previousStrategyName) {
--- a/lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
+++ b/lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\ODM\PHPCR\Tools\Helper;
 
+use Doctrine\Common\Util\ClassUtils;
 use Doctrine\ODM\PHPCR\Exception\InvalidArgumentException;
 use Doctrine\ODM\PHPCR\PHPCRExceptionInterface;
 use Doctrine\ODM\PHPCR\Translation\Translation;
@@ -38,6 +39,11 @@ class TranslationConverter
     private $batchSize;
 
     /**
+     * @var array
+     */
+    private $notices = array();
+
+    /**
      * @param DocumentManagerInterface $dm
      * @param int                      $batchSize
      */
@@ -54,8 +60,9 @@ class TranslationConverter
      * PHPCR session *after each batch*. Calling flush on the document manager
      * *is not enough*.
      *
-     * When un-translating, the current locale and language fallback is used.
-     * When translating, the new properties are copied into all languages.
+     * When translating, the new properties are copied into the languages
+     * specified in $locales. When un-translating, the current locale and
+     * language fallback is used, and $locales is ignored.
      *
      * To convert all fields into a translation, you can pass an empty array
      * for $fields and the information is read from the metadata. The fields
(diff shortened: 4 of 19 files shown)


lib/Doctrine/ODM/PHPCR/Query/Query.php
```diff
@@ -180,7 +188,7 @@ public function execute($parameters = null, $hydrationMode = null)
                 $data = $this->query->execute();
                 break;
             case self::HYDRATE_DOCUMENT:
-                $data = $this->dm->getDocumentsByPhpcrQuery($this->query, $this->documentClass, $this->primaryAlias);
+                $data = $this->dm->getDocumentsByPhpcrQuery($this->query, $this->documentClass);
```
dbu: last argument was unused

lib/Doctrine/ODM/PHPCR/Translation/TranslationStrategy/AttributeTranslationStrategy.php
```diff
@@ -163,7 +163,14 @@ public function removeTranslation($document, NodeInterface $node, ClassMetadata
      */
     public function removeAllTranslations($document, NodeInterface $node, ClassMetadata $metadata)
     {
-        // Do nothing: if the node is removed then all it's translated properties will be removed
+        foreach ($this->getLocalesFor($document, $node, $metadata) as $locale) {
```
dbu: we now need this for the translation migrations

tests/Doctrine/Tests/ODM/PHPCR/Functional/Tools/Helper/TranslationConverterTest.php
```diff
@@ -0,0 +1,445 @@
...
+            'Move to untranslated'
+        );
+        $this->session->save();
+
+        $this->assertFalse($this->converter->convert($class, null, array($field)));
+        $this->session->save();
+        $this->dm->clear();
+
+        $this->dm = $this->createDocumentManager();
+        $this->dm->setLocaleChooserStrategy(new LocaleChooser($this->localePrefs, 'en'));
```
dbu: at this point the metadata is pretty much f*ked up. we might try to restore it. but the doc should tell to not even try to use the dm anymore after a migration.

lsmith77:
very nice! .. should be put this into 1.3.0 ? which would mean another RC

lib/Doctrine/ODM/PHPCR/Query/Query.php
dantleech: should we also remove the member property then?

dantleech: Is it really unused? It is set in the constructor and is indeed the third argument of `getDocumentsByPhpcrQuery`.

tests/Doctrine/Tests/ODM/PHPCR/Functional/Tools/Helper/TranslationConverterTest.php
```diff
@@ -0,0 +1,445 @@
...
+        $this->assertEquals('Lorem ipsum...', $commentDoc->getText());
+
+        $this->dm->clear();
+
+        $commentDoc = $this->dm->find(null, '/functional/convert');
+        $this->assertInstanceof($class, $commentDoc);
+        $this->assertEquals('Lorem ipsum...', $commentDoc->getText());
+    }
+
+
```
dantleech: dd

tests/Doctrine/Tests/ODM/PHPCR/Functional/Tools/Helper/TranslationConverterTest.php
dantleech: would need to read the code more closely, but why is it broken? I guess we override the metadata to transition the old translation strategy? But then would it be so difficult to restore the original metadata?

saying that, as this would be executed as a command most of the time, it shouldn't be an issue. Maybe this would also affect migrations, but I would expect them to be done at the PHPCR level.

dbu:
well i think it would be nice to get it in. lets see if we can wrap things up this week and otherwise move on. the impact on the existing code base is very small, its mainly new classes so i am not too afraid of disrupting something

lsmith77:
agreed

lib/Doctrine/ODM/PHPCR/Query/Query.php
dbu: ooops. DocumentManagerInterface only has 2 parameters, but the actual method has 3. also the dm decorator does not have that 3rd parameter, so it will drop it when used.

we should probably fix that interface then, even though it means potentially breaking custom DMs . but dropping parameters is worse than such a break imo

tests/Doctrine/Tests/ODM/PHPCR/Functional/Tools/Helper/TranslationConverterTest.php
dbu: i tried to restore information but seem to be missing something.

in a migration process with a custom command using the helper, the metadata mixup could matter so i would love to fix it. anybody sees what is happening? might also be something with caching inside the uow?

lib/Doctrine/ODM/PHPCR/Query/Query.php
dbu: see #656

dbu:
it seems to me the only reason for failure is now due to https://github.com/phpcr/phpcr-utils/pull/157

funnily, dbal seems to work with the hyphens, only jackrabbit complains. maybe we miss something in the dbal implementation but maybe its a lucene implementation detail that requires the brackets. we might want to add this check to dbal to be consistent, but that is not in the scope of this PR

composer.json
```diff
@@ -18,7 +18,7 @@
         "doctrine/data-fixtures": "~1.0,>=1.0.0",
         "phpcr/phpcr": "~2.1.1",
         "phpcr/phpcr-implementation": "~2.1.0",
-        "phpcr/phpcr-utils": "~1.2.7|~1.3",
+        "phpcr/phpcr-utils": "dev-always-escape-propertyname as 1.2.7",
```
dbu: don't forget to update once phpcr-utils is merged and tagged.

dbu:
added documentation in https://github.com/doctrine/phpcr-odm-documentation/pull/77

this is now ready. we should update the odm bundle once its merged to also support the command inside symfony

lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentConvertTranslationCommand.php
```diff
@@ -0,0 +1,109 @@
...
+            );
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        /** @var $session SessionInterface */
+        $session = $this->getHelper('phpcr')->getSession();
```
lsmith77: should we make this command interactive and by default require confirmation? we could even list exactly list what fields are going to be converted

dbu: that sounds like a good idea. and showing what will happen would be great.

i can add a --force parameter and otherwise ask for confirmation. for
the rest, i don't see that i have time to do that soon enough. i don't
want to delay the release. the --force is a BC break, providing more
information is possible later as well. we would need to refactor the
class that does the migration to be able to explain its decisions
without executing them yet.

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
```diff
@@ -0,0 +1,228 @@
...
+     *                                           field not translated anymore
+     *
+     * @return boolean true if there are more documents to convert and this method needs to be
+     *                      called again.
+     *
+     * @throws PHPCRExceptionInterface if the document can not be found.
+     */
+    public function convert(
+        $class,
+        $previousStrategyName = null,
```
liquorvicar: As I understand it, this takes the string value representing the strategy (i.e. "attribute" or "child" unless a custom strategy has been implemented). Would it be better if these were constants? Or perhaps that this method accepted the class name?

liquorvicar:
So this may not be an issue with the Converter per se but it's something we've run into and I thought it was worth raising here (there may be an easy way round this or I may be misunderstanding something).

We have an existing site built on phpcr-odm (via Symfony-CMF) and we now have a requirement to make it translatable. The Converter works fine so far and we have written a migration to convert various pages to be translatable and preserve the existing, untranslatable content. I am now trying to write the "down" for the migration and have hit a block. My down is currently just this:
```
$converter = new TranslationConverter($this->dm);

foreach ($this->pagesToConvert as $pageToConvert) {
    $converter->convert($pageToConvert, 'attribute');
}

$this->dm->getPhpcrSession()->save();
$this->dm->flush();
```
However when we run it we get an error: "Previous and current strategy are the same. Document is currently at attribute". It looks like the Converter is reading the current strategy for the Document as per the config (we have set this to be "attribute" globally via config.yml). How would we get round this? Do we need to set the translator manually on the Document before we run the Converter in the down() method?

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
dbu: you understood correctly. i updated the docblock to be explicit.

regarding constants: i agree that we should provide constants for the default strategies. currently, their names are simply written in https://github.com/doctrine/phpcr-odm/blob/translation-conversion/lib/Doctrine/ODM/PHPCR/DocumentManager.php#L129-L130

if you want to do a pull request to provide the constants and replace string occurences with the constants, please do. not sure where the best location is though - the interface seems a bit wrong, maybe each strategy implementation? when you use the constant you do know about the strategy...

dbu:
good point. i think your safest "down" at that point is a restore of the repository - i would not trust 100% that things will get good if something messed up at that point.

but i added another optional parameter to specify the current strategy as well. i am not 100% sure if all the logic works out as it should, but i run out of time here. can you try it and if it does not work, provide a pull request against this branch that would fix it?

dbu:
@liquorvicar can you check if this change helps for your "down"? sorry for the rush, but this is one of the last blockers for tagging an RC of the next minor version, so i want to merge this asap.

liquorvicar:
@dbu No, unfortunately that hasn't helped. The command we're using for the down now is:
```
    $converter = new TranslationConverter($this->dm);
    $converter->convert(BasicPage::class, ['title', 'intro', 'strapline', 'label'], 'attribute', new NonTranslatedStrategy($this->dm));
```
I have pasted below the props for one of our documents after the successful up and then after the down so you can see what's going on (I tried attaching them but I don't have the right permissions). Looks like all the translatable fields have been dropped and nothings been added back in. Might be better to follow your suggestion of taking a dump of the repository and restoring it as the down. Is there an easy/recommended way to do that in code?

I will see if I can work out how to fix it and submit a PR but my knowledge of the internals here is pretty sketchy...

Document props after converting to translatable:
```
<?xml version="1.0" encoding="UTF-8"?>
<sv:node xmlns:mix="http://www.jcp.org/jcr/mix/1.0" xmlns:nt="http://www.jcp.org/jcr/nt/1.0"
         xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:jcr="http://www.jcp.org/jcr/1.0"
         xmlns:sv="http://www.jcp.org/jcr/sv/1.0" xmlns:rep="internal">
    <sv:property sv:name="jcr:primaryType" sv:type="Name" sv:multi-valued="0">
        <sv:value length="15">nt:unstructured</sv:value>
    </sv:property>
    <sv:property sv:name="jcr:mixinTypes" sv:type="Name" sv:multi-valued="1">
        <sv:value length="13">phpcr:managed</sv:value>
        <sv:value length="17">mix:referenceable</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr:class" sv:type="String" sv:multi-valued="0">
        <sv:value length="42">Autologic\ContentBundle\Document\BasicPage</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr:classparents" sv:type="String" sv:multi-valued="1">
        <sv:value length="31">Symfony\Component\Routing\Route</sv:value>
        <sv:value length="44">Symfony\Cmf\Bundle\RoutingBundle\Model\Route</sv:value>
        <sv:value length="53">Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Phpcr\Route</sv:value>
        <sv:value length="54">Symfony\Cmf\Bundle\SimpleCmsBundle\Doctrine\Phpcr\Page</sv:value>
    </sv:property>
    <sv:property sv:name="jcr:uuid" sv:type="String" sv:multi-valued="0">
        <sv:value length="36">db06c2e7-e934-49b1-8beb-9dd9a6e155e3</sv:value>
    </sv:property>
    <sv:property sv:name="host" sv:type="String" sv:multi-valued="0">
        <sv:value length="0"></sv:value>
    </sv:property>
    <sv:property sv:name="schemes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="methods" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="defaults" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="requirements" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="options" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="publishable" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="extras" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="display" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="displayChildren" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="attributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="childrenAttributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="linkAttributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="labelAttributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="template" sv:type="String" sv:multi-valued="0">
        <sv:value length="7">contact</sv:value>
    </sv:property>
    <sv:property sv:name="visible_for_sitemap" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:hy_AM-title" sv:type="String" sv:multi-valued="0">
        <sv:value length="10">Contact Us</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:hy_AM-label" sv:type="String" sv:multi-valued="0">
        <sv:value length="0"></sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:hy_AM-intro" sv:type="String" sv:multi-valued="0">
        <sv:value length="186">&lt;p&gt;We're always happy to hear from you. To get in touch with Autologic, please use
            the form below or use the contact details for any of our offices found at the bottom of this page.&lt;/p&gt;
        </sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:hy_AMnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_US-title" sv:type="String" sv:multi-valued="0">
        <sv:value length="15">Howdy, partner!</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_US-label" sv:type="String" sv:multi-valued="0">
        <sv:value length="0"></sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_US-intro" sv:type="String" sv:multi-valued="0">
        <sv:value length="186">&lt;p&gt;We're always happy to hear from you. To get in touch with Autologic, please use
            the form below or use the contact details for any of our offices found at the bottom of this page.&lt;/p&gt;
        </sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_USnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_GB-title" sv:type="String" sv:multi-valued="0">
        <sv:value length="13">Good Day, sir</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_GB-label" sv:type="String" sv:multi-valued="0">
        <sv:value length="0"></sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_GB-intro" sv:type="String" sv:multi-valued="0">
        <sv:value length="186">&lt;p&gt;We're always happy to hear from you. To get in touch with Autologic, please use
            the form below or use the contact details for any of our offices found at the bottom of this page.&lt;/p&gt;
        </sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_GBnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:fr_FR-title" sv:type="String" sv:multi-valued="0">
        <sv:value length="15">Contactez-nous!</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:fr_FR-label" sv:type="String" sv:multi-valued="0">
        <sv:value length="0"></sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:fr_FR-intro" sv:type="String" sv:multi-valued="0">
        <sv:value length="186">&lt;p&gt;We're always happy to hear from you. To get in touch with Autologic, please use
            the form below or use the contact details for any of our offices found at the bottom of this page.&lt;/p&gt;
        </sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:fr_FRnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:de_DE-title" sv:type="String" sv:multi-valued="0">
        <sv:value length="12">Guten Morgen</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:de_DE-label" sv:type="String" sv:multi-valued="0">
        <sv:value length="0"></sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:de_DE-intro" sv:type="String" sv:multi-valued="0">
        <sv:value length="186">&lt;p&gt;We're always happy to hear from you. To get in touch with Autologic, please use
            the form below or use the contact details for any of our offices found at the bottom of this page.&lt;/p&gt;
        </sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:de_DEnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
</sv:node>
```

Document props after running the "down" migration:
```
<?xml version="1.0" encoding="UTF-8"?>
<sv:node xmlns:mix="http://www.jcp.org/jcr/mix/1.0" xmlns:nt="http://www.jcp.org/jcr/nt/1.0"
         xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:jcr="http://www.jcp.org/jcr/1.0"
         xmlns:sv="http://www.jcp.org/jcr/sv/1.0" xmlns:rep="internal">
    <sv:property sv:name="jcr:primaryType" sv:type="Name" sv:multi-valued="0">
        <sv:value length="15">nt:unstructured</sv:value>
    </sv:property>
    <sv:property sv:name="jcr:mixinTypes" sv:type="Name" sv:multi-valued="1">
        <sv:value length="13">phpcr:managed</sv:value>
        <sv:value length="17">mix:referenceable</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr:class" sv:type="String" sv:multi-valued="0">
        <sv:value length="42">Autologic\ContentBundle\Document\BasicPage</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr:classparents" sv:type="String" sv:multi-valued="1">
        <sv:value length="31">Symfony\Component\Routing\Route</sv:value>
        <sv:value length="44">Symfony\Cmf\Bundle\RoutingBundle\Model\Route</sv:value>
        <sv:value length="53">Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Phpcr\Route</sv:value>
        <sv:value length="54">Symfony\Cmf\Bundle\SimpleCmsBundle\Doctrine\Phpcr\Page</sv:value>
    </sv:property>
    <sv:property sv:name="jcr:uuid" sv:type="String" sv:multi-valued="0">
        <sv:value length="36">db06c2e7-e934-49b1-8beb-9dd9a6e155e3</sv:value>
    </sv:property>
    <sv:property sv:name="host" sv:type="String" sv:multi-valued="0">
        <sv:value length="0"></sv:value>
    </sv:property>
    <sv:property sv:name="schemes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="methods" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="defaults" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="requirements" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="options" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="publishable" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="extras" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="display" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="displayChildren" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="attributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="childrenAttributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="linkAttributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="labelAttributes" sv:type="String" sv:multi-valued="1"/>
    <sv:property sv:name="template" sv:type="String" sv:multi-valued="0">
        <sv:value length="7">contact</sv:value>
    </sv:property>
    <sv:property sv:name="visible_for_sitemap" sv:type="Boolean" sv:multi-valued="0">
        <sv:value length="1">1</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:hy_AMnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_USnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:en_GBnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:fr_FRnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
    <sv:property sv:name="phpcr_locale:de_DEnullfields" sv:type="String" sv:multi-valued="1">
        <sv:value length="4">body</sv:value>
        <sv:value length="9">strapline</sv:value>
    </sv:property>
</sv:node>
 ```

**Comment by dbu** [comment]
> @dbu <https://github.com/dbu> No, unfortunately that hasn't helped. The
> command we're using for the down now is:
> 
> |$converter = new TranslationConverter($this->dm);
> $converter->convert(BasicPage::class, ['title', 'intro', 'strapline',
> 'label'], 'attribute', new NonTranslatedStrategy($this->dm)); |

hm, that last argument should be the name, not an object. but how to
distinguish between null = no translation and null = not specified...
maybe we could use boolean `false`?

> back in. Might be better to follow your suggestion of taking a dump of
> the repository and restoring it as the down. Is there an
> easy/recommended way to do that in code?

tried to find the right documentation page but there is not really one.
but when running app/console, you have doctrine:phpcr:workspace:export
and doctrine:phpcr:workspace:import. the --help gives you information
how they work.

> I will see if I can work out how to fix it and submit a PR but my
> knowledge of the internals here is pretty sketchy...

what to do for now? @lsmith77 do we just remove the target strategy
optional parameter until somebody has time to make this work correctly
and with tests?

**Comment by liquorvicar** [comment]
> hm, that last argument should be the name, not an object. but how to distinguish between null = no 
> translation and null = not specified... maybe we could use boolean `false`?

Yes, I hit that issue which is why I thought I'd pass in the object. Maybe the string "none" would be better than false (so it's consistent with the other strategies being strings).

> tried to find the right documentation page but there is not really one. but when running app/console, you have doctrine:phpcr:workspace:export and doctrine:phpcr:workspace:import. the --help gives you information how they work.

Thanks, that looks like it gives me what I need.

**Comment by dbu** [comment]
i will remove the current strategy parameter for now and switch from
null to none for the strategy name to be explicit. then we are ready to
add this feature in the future.

**Comment by dbu** [comment]
@lsmith77 updated. time for final review?

**Comment by liquorvicar** [comment]
I have come across one issue with our conversions. Not sure if it's a bug in the converter or a problem due to the way we are structuring our content. We have a number of PHP classes implementing various types of Document. Some of these extend base classes (which themselves have no concrete nodes). Without realising, we added one of these base classes to the list of classes to convert, and crucially, it was in the list *before* one of the child classes that extended it. The upshot was that the nodes using the child class that was converted *after* the base class that child extended did not get converted correctly (although oddly the fields in the parent class were not even converted).

Obviously, it was our mistake to include the parent class that did not have any nodes directly using it, but it does feel like the converter might have handled this better.

Any thoughts? Shall I create a separate issue for this? Or add a comment/some documentation to explain? Of course, we may just be doing things in a way that was never intended with SymfonyCMF/phpcr but if we've done it, chances are someone else will trip over the same problem...

**Comment by dbu** [comment]
glad you noticed. yeah of course, this is a problem. it seems a valid
use case, i do that myself as well. i guess we should add an option
`--strict` that you can set to `false` and usually only convert when the
class name is a strict match. if you know all child classes need no
conversion, you can run the command with `--strict=false` but by default
only the exact class is converted.

**Comment by liquorvicar** [comment]
So the default would be to convert only the named classes passed in to the converter (strict: true) but you could override that if you wanted to convert all instances of a parent class instead (strict: false). Yes, I think that would make sense to me. Pretty sure that would have prevented us tripping over this issue ;-)

**Comment by dbu** [comment]
i will have a look if i can output warnings on the cli if things had to
be skipped because of the strict flag. otherwise it will be hard to
figure out why some have not been converted

File: lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php [hunk 8]
```diff
@@ -0,0 +1,256 @@
...
+     * The current strategy is read from the document metadata.
+     *
+     * @param string        $class                FQN of the document class
+     * @param array         $fields               List of fields to convert. Required when making a
+     *                                            field not translated anymore
+     * @param string        $previousStrategyName Name of previous strategy or "none" if field was not
+     *                                            previously translated
+     * @param array|boolean $strict               Whether to convert descending classes of $class as well.
+     *                                            Pass an empty array if you want to collect which documents
+     *                                            had to be skipped. Pass `false` to convert descendants.
```
dbu: not really happy with the parameter multi-functionality (boolean vs array for return value) but returning an array with a flag for more in the loop and for notices seems just as ugly.

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        // fake metadata for previous
+        $previousMeta = clone $currentMeta;
+        $previousMeta->translator = NonTranslatedStrategy::NAME === $previousStrategyName ? null : $previousStrategyName;
+        // even when previously not translated, we use translatableFields for the NonTranslatedStrategy
+        $previousMeta->translatableFields = $fields;
+
+        foreach ($documents as $document) {
+            if (false !== $strict && ClassUtils::getClass($document) !== $class) {
```
dbu: this is the new strict functionality

tests/Doctrine/Tests/Models/Translation/DatedComment.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Doctrine\Tests\Models\Translation;
+
+use Doctrine\ODM\PHPCR\Mapping\Annotations as PHPCRODM;
+
+/**
+ * @PHPCRODM\Document(translator="attribute")
```
dbu: i thought we fixed that the attribute from the parent is inherited. but without this i get an error that "text" is translated but there is no translator strategy

tests/Doctrine/Tests/ODM/PHPCR/Functional/Tools/Helper/TranslationConverterTest.php
```diff
@@ -0,0 +1,516 @@
...
+        $comment->setProperty('phpcr:classparents', array($parentClass));
+        $this->session->save();
+
+        $strict = false;
+        $this->assertFalse($this->converter->convert($parentClass, array(), NonTranslatedStrategy::NAME, $strict));
+
+        $this->session->save();
+
+        // TODO the document is not correctly loaded, the text field stays null.
+        var_dump($comment->getPropertiesValues());
```
dbu: argh. did not find the mistake yet.

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
```diff
@@ -0,0 +1,256 @@
...
+            if (NonTranslatedStrategy::NAME === $currentStrategyName) {
+                throw new InvalidArgumentException('To untranslate a document, you need to specify the fields that where previously translated');
+            }
+            $fields = $currentMeta->translatableFields;
+        }
+
+        // trick query into using the previous strategy
+        $currentMeta->translator = NonTranslatedStrategy::NAME === $previousStrategyName ? null : $previousStrategyName;
+
+        // TODO do we need to change strategy on ancestors as well?
```
dbu: ok, this is only a problem in non-strict mode. 
when we convert a child of $class, the metadata of that child will be used, not the metadata of the parent. hence we would need to change the strategy of the child class metadata. how to do that? do we register a metadata load event listener on the fly? or re-load documents with fixed metadata if they are not strictly the same class?

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
lsmith77: what do you mean "for return value"? because its pass by ref? another option could be to store this information in a class property so that no pass by ref is needed ..

BTW in that case one could also choose to move the return value into a class property so that one could return the array of class names from the method .. 

but yeah it feels a bit like choosing a poison

dbu: ah, good idea! i just add a getLastNotices method that provides this information. much better than this parameter by reference mess.

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
lsmith77: not sure under which conditions we trigger the load event .. I would suspect only on first load but what if it was previously loaded for some reason? maybe lets just keep it simple and only provide strict mode for now?

dbu: sounds reasonable. it will be the default anyways. but then i remove the
option and command flag again, and just output warnings about things not
being transformed because we do strict matches. users just need to rerun
the command for child classes, even if theoretically it would be the
same thing.

dbu:
ok, i think we are getting there. added the locales option and removed strict, we are just always strict. refactored code to be cleaner.

lsmith77:
```
1) Doctrine\Tests\ODM\PHPCR\Tools\Command\DocumentConvertTranslationCommandTest::testCommand
Expectation failed for method name is equal to <string:convert> when invoked 1 time(s)
Parameter 2 for invocation Doctrine\ODM\PHPCR\Tools\Helper\TranslationConverter::convert('Document\MyClass', Array (), Array (), 'none') does not match expected value.
Array () does not match expected type "string".
```

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
```diff
@@ -0,0 +1,275 @@
...
+     * @param int                      $batchSize
+     */
+    public function __construct(DocumentManagerInterface $dm, $batchSize = 200)
+    {
+        $this->dm = $dm;
+        $this->batchSize = $batchSize;
+    }
+
+    /**
+     * Migrate content into the new translation format and remove the old properties.
```
lsmith77: phpdoc should maybe reference ``getLastNotices()`` as well

lsmith77:
aside from the build failure it looks good to me!

lib/Doctrine/ODM/PHPCR/Tools/Helper/TranslationConverter.php
dbu: added an `@see` - the method is mentioned in the text already

PR #656, opened 2015-08-27
fixing DocumentManagerInterface

dbu:
fix bugs revealed in #655 

we seem to miss a mess detector check to tell when the interface and an implementation do not match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+* **2015-08-27** **BC break** Missing parameter in `DocumentManagerInterface::getDocumentsByPhpcrQuery`
+                 added. If you have your own implementation that is not based
+                 on DocumentManagerDecorator, you will need to adjust that method.
+
 1.3.0-rc2
 ---------
 
--- a/lib/Doctrine/ODM/PHPCR/Decorator/DocumentManagerDecorator.php
+++ b/lib/Doctrine/ODM/PHPCR/Decorator/DocumentManagerDecorator.php
@@ -225,9 +225,9 @@ abstract class DocumentManagerDecorator extends ObjectManagerDecorator implement
     /**
      * {@inheritDoc}
      */
-    public function getDocumentsByPhpcrQuery(QueryInterface $query, $className = null)
+    public function getDocumentsByPhpcrQuery(QueryInterface $query, $className = null, $primarySelector = null)
     {
-        return $this->wrapped->getDocumentsByPhpcrQuery($query, $className);
+        return $this->wrapped->getDocumentsByPhpcrQuery($query, $className, $primarySelector);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+* **2015-08-27** **BC break** Missing parameter in `DocumentManagerInterface::getDocumentsByPhpcrQuery`
+                 added. If you have your own implementation that is not based
+                 on DocumentManagerDecorator, you will need to adjust that method.
```
dbu: a typical bugfix BC break.

i think this is really unlikely to hit people, as they probably where using the decorator and then unless they overwrite the getDocumentsByPhpcrQuery method will not notice the change.

lib/Doctrine/ODM/PHPCR/Decorator/DocumentManagerDecorator.php
```diff
@@ -235,7 +235,7 @@ public function getDocumentsByPhpcrQuery(QueryInterface $query, $className = nul
      */
     public function persist($document)
     {
-        return $this->wrapped->persist($document);
+        $this->wrapped->persist($document);
```
dbu: declared to return `void`, lets not trick people into thinking this could return something.

lsmith77:
+1

lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php
```diff
@@ -489,6 +490,24 @@ public function findVersionByName($className, $id, $versionName);
...
+     *
+     * @param object|array|null $document optionally limit to a specific
+     *      document or an array of documents
+     *
+     * @return void
+     *
+     * @throws InvalidArgumentException if $document is neither null nor a
+     *      document or an array of documents
+     */
+    public function flush($document = null);
```
lsmith77: this causes on PHP 5.3.3 the following error:
```
PHP Fatal error:  Can't inherit abstract function Doctrine\Common\Persistence\ObjectManager::flush() (previously declared abstract in Doctrine\ODM\PHPCR\DocumentManagerInterface) in /home/travis/build/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php on line 39
```

dbu: time to drop php 5.3.3 support? or do you want to not declare this? it will mean that decorators risk to destroy individual flush

lsmith77: does it work with 5.3.9 ?

dbu: i just adapted the .travis.yml to 5.3 instead of 5.3.3 - lets see. if its good, we should raise the php version in composer.json to 5.3.9 (i won't have a computer with me for the next 3 weeks...)

dbu: yay, looks green. and it seems 5.3.9 we need: https://3v4l.org/6SnfG

dbu:
bumped to 5.3.9 - please merge if its green.

Issue #659, opened 2015-09-14
Changing the name of a node results in deletion not move


uwej711:
I found this while trying to rename a route (from the RoutingBundle) with setName, the cause is a bit difficult to describe ...

First you need a parent with mapped children, say /cms/routes/zn, the child /cms/routes/zn/page-1 will be renamed to page-2 with setName.

The UnitOfWork correctly detects the move of /cms/routes/zn/page-1 to /cms/routes/zn/page-2. But it also creates a change for /cms/routes/zn because it adds an empty reordering in https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/UnitOfWork.php#L1422 or https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/UnitOfWork.php#L1424. the resulting executeUpdates for /cms/routes/zn triggers another computeChangeset for that node in https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/UnitOfWork.php#L2482, which then results in a deletion of /cms/routes/zn/page-1.

Only adding the reodering if the array in https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/UnitOfWork.php#L1420 has count > 0 solves that very special case for me but if your parent node has other changes it will break anyway.

Hope this is somehow clear ...

Cheers
Uwe

dbu:
i guess we should not add the empty reordering, it makes no sense. but for when other changes do exist, we have to sort out the ordering issue then.

uwej711:
The setup to demonstrate this seems a bit more complicated than I thought ...

dantleech:
So this UOW is moving the node, then recalculating the changeset, and then registering a delete for the "missing" node and then deleting it because `executeRemoves` is after `executeUpdates` ?

- Could we calculate the delete based on UUID instead of node names? (only if we were to force UUIDs in a next major version).

uwej711:
This is a bit more complicated, see the test in #661. 
Because there is an event listener computeChangeset is called twice but the second computation messes up the child renaming. In #661 I restricted this changeset computation to fields only bit that leaves us with a failing test that want's to revert a child reorder in the preUpdate event.

uwej711:
Just bumped into that again. I think I will test my change in #661 in production now ...

dbu:
if you find a way to get the build green on #661 and have an improvement
[...]

PR #661, opened 2015-09-18
add failing test for #659

uwej711:
so there a two conditions to make that test fail
1. add an preUpdate Listener to the parent class
2. make sure the parents ChildrenCollection is initialized
without either one of these conditions the test passes!

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -1441,7 +1441,7 @@ class UnitOfWork
      * @param ClassMetadata $class
      * @param object        $document
      */
-    public function computeChangeSet(ClassMetadata $class, $document)
+    public function computeChangeSet(ClassMetadata $class, $document, $fieldsOnly = false)
     {
         if ($document instanceof Proxy && !$document->__isInitialized()) {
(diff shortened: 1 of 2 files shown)


uwej711:
when we restrict preUpdate to fields only, there remains one failing test (see travis). Not sure, if this is the way to go. But I see no other simple solution for #659. I think it would be safer anyway to restrict the possible changes to a documents fields as there a no other changesets computed after the event. But that currently leaves us with no way to reset the order of children (maybe this should be allowed).

uwej711:
I think this needs some discussion ...

dbu:
events during the flush seem to be a really tricky beast. do you see a way to allow field updates and reorderings, but no renaming? or could we store the renaming operation in a way that the second changeset calculation can check if something was renamed and not get confused?

could we look at object identity rather than paths? uuid would only be something for 2.0 as we would change everything to have a uuid.

uwej711:
I can try to look into the reordering but I fear this will not be an easy thing to fix.

dbu:
what are our other options? limiting event listener to update properties?

uwej711:
That's already done here, with the one test about reverting the reordering of the children in preUpdate failing, i.e. that would be a bc break of some kind.

PR #662, opened 2015-09-30
failing test for calling same with empty string

uwej711:
this simple tests fails when using jackrabbit - not sure what the expected behaviour is, let's have a look at the spec?

### DIFF
--- a/tests/Doctrine/Tests/ODM/PHPCR/Functional/QueryBuilderTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Functional/QueryBuilderTest.php
@@ -399,7 +399,21 @@ class QueryBuilderTest extends PHPCRFunctionalTestCase
     }
 
     /**
-     * @expectedException InvalidArgumentException 
+     * @depends testFrom
+     */
+    public function testSameEmpty()
+    {
+        $qb = $this->createQb();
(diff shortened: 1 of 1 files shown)


dbu:
looking at http://www.day.com/specs/jcr/2.0/6_Query.html i am not exactly sure what we do. is this the same node thing? then the selector for the target node is not optional. i think what needs to be improved:

* validate in phpcr-odm query builder (and phpcr qom maybe?) to throw a specific exception if the target is empty
* validate in jackalope-doctrine-dbal when the target is empty

semantically it makes no sense to filter for a node to be the same as *nothing*.

uwej711:
from the spec:

"The query is invalid if:
selectorName is not the name of a selector in the query, or
path is not a syntactically valid absolute path (see §3.3.4 Lexical Path Grammar). Note, however, that if path is syntactically valid but does not identify a node in the workspace (or the node is not visible to this session, because of access control constraints), the query is valid but the constraint is not satisfied."

So lets check if the empty string is a valid path?

dbu:
> So lets check if the empty string is a valid path?

empty string is not a valid path.

lsmith77:
so it seems like we should add some validation logic for this to ensure that we fail with a sensible exception message

Issue #657, opened 2015-08-29
Regression when persisting object ID and changing the locale


dantleech:
A test has been [added](https://github.com/symfony-cmf/RoutingAutoBundle/pull/159/files#diff-d35d0adbd9705b4b3930c46cb17ba224R273) in the RoutingAutoBundle which seems to reveal a regression introduced in 83863d7e460c112aff1497abc4fdfb9c7b93cde2.

Failure is as follows

````
1) Symfony\Cmf\Bundle\RoutingAutoBundle\Tests\Functional\EventListener\AutoRouteListenerTest::testMultilangArticleRemainsSameLocale
Undefined index: path

/home/daniel/www/symfony-cmf/RoutingAutoBundle/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:40
/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/UnitOfWork.php:1604
/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/UnitOfWork.php:1123
/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/UnitOfWork.php:2157
/home/daniel/www/doctrine/phpcr-odm/lib/Doctrine/ODM/PHPCR/DocumentManager.php:949
/home/daniel/www/symfony-cmf/RoutingAutoBundle/Tests/Functional/EventListener/AutoRouteListenerTest.php:288
````

/cc @lsmith77 @WouterJ

WouterJ:
I suspect it is caused by this PR: https://github.com/doctrine/phpcr-odm/pull/555 (which was merged in 1.2.0-RC4, the first version failing). So this has to be fixed in both 1.2.x and master.

WouterJ:
Btw, the problem probably is not in the fact that path is the Id, but that the Id is one of the types that is not translatable. Only the translatable fields seem to exists at the line of the error.

dantleech:
ok, I know what the problem is.

When the changeset if calculated the "isNew" flag is determined by the presence of the object ID in the `$originalData` array. If [isNew` is false](https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/UnitOfWork.php#L1459) then we expect all the fields to be present in the `$originalData`.

[...]

PR #665, opened 2015-10-09
Failing test for issue #657 (calling find before flush)

dantleech:
Failing test for issue #657

### DIFF
--- a/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/DocumentManagerTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/DocumentManagerTest.php
@@ -881,4 +881,31 @@ class DocumentManagerTest extends PHPCRFunctionalTestCase
 
         $this->assertEquals('Guten tag', $trans->topic);
     }
+
+    /**
+     * It should be able to find a bound translation before flush.
+     */
+    public function testRetainTranslation()
+    {
+        $article = new Article;
+        $article->id = '/functional/article-1';
+        $article->text = 'Foo';
+        $this->dm->persist($article);
+
+        $article->topic = 'Hello everybody!';
+        $this->dm->bindTranslation($article, 'en');
+
+        $article->topic = 'Bonjour le monde!';
+        $this->dm->bindTranslation($article, 'fr');
+        $articleInstance = $this->dm->findTranslation(get_class($article), '/functional/article-1', 'en');
+
+        // Fails - returns "foo"
+        // $this->assertEquals('Hello everybody!', $articleInstance->text);
+
+        // Fails, could not find /functional/article-1
+        $this->dm->flush();
+        $this->dm->clear();
+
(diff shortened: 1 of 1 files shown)


dantleech:
Note that this does not currently address the same problem which this case was designed to reproduce (in [routing auto](https://github.com/symfony-cmf/RoutingAutoBundle/pull/159/files#diff-d35d0adbd9705b4b3930c46cb17ba224R273). In that issue `path` was undefined.

I tried to fix this in a few ways yesterday, but I think there may be more than one issue that needs to be overcome.

tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/DocumentManagerTest.php
```diff
@@ -881,4 +881,31 @@ public function testAdditionalFindCallsDoNotRefresh()
...
+        $article->id = '/functional/article-1';
+        $article->text = 'Foo';
+        $this->dm->persist($article);
+
+        $article->topic = 'Hello everybody!';
+        $this->dm->bindTranslation($article, 'en');
+
+        $article->topic = 'Bonjour le monde!';
+        $this->dm->bindTranslation($article, 'fr');
+        $articleInstance = $this->dm->findTranslation(get_class($article), '/functional/article-1', 'en');
```
dbu: this will be the same object as $article. not sure if we should assign in this case. if we have no test about that, maybe assertSame($article, $articleInstance).

dbu:
i agree that this should work like the test describes it

tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/DocumentManagerTest.php
dantleech: Yeah, I was trying to illustrate the problrem, but yeah, pointlesss.

On 9 October 2015 16:45:53 BST, David Buchmann <notifications@github.com> wrote:
>> +    /**
>> +     * It should be able to find a bound translation before flush.
>> +     */
>> +    public function testRetainTranslation()
>> +    {
>> +        $article = new Article;
>> +        $article->id = '/functional/article-1';
>> +        $article->text = 'Foo';
>> +        $this->dm->persist($article);
>> +
>> +        $article->topic = 'Hello everybody!';
>> +        $this->dm->bindTranslation($article, 'en');
>> +
>> +        $article->topic = 'Bonjour le monde!';
>> +        $this->dm->bindTranslation($article, 'fr');
>> +        $articleInstance =
>$this->dm->findTranslation(get_class($article),
>'/functional/article-1', 'en');
>
>this will be the same object as $article. not sure if we should assign
>in this case. if we have no test about that, maybe assertSame($article,
>$articleInstance).
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/doctrine/phpcr-odm/pull/665/files#r41645115

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Issue #658, opened 2015-09-06
Query builder seems unable to produce a fully working left join


sarcher:
I am attempting to use a `LEFT JOIN` behavior on child documents, but it seems to be impossible based on what I've seen.

I have the following situation:

* Tree of documents of the class `Product`
* Documents _may_ have one or more children of the class `ProductDetail`
* Attempting to design a search query which includes data from these details

Take the following query builder as an example:

```php
$searchTerm = '12345';

$query = $queryBuilder
    ->fromDocument('Path\To\Product', 'p')
    ->addJoinLeftOuter()
        ->right()->document('Path\To\ProductDetail', 'pd')->end()
        ->condition()->child('pd', 'p')->end()
    ->end();
    ->andWhere()
        ->orX()
            ->like()->localName('p')->literal('%' . $searchTerm . '%')->end();
            ->andX()
                ->eq()->field('pd.type')->literal('title')->end()
                ->fullTextSearch('pd.content', $searchTerm)
            ->end()
        ->end()
    ->end()
    ->getQuery();
```

Now, PHPCR-ODM produces a JCR-SQL2 query like:

```sql
SELECT * 
FROM [nt:unstructured] AS p 
LEFT OUTER JOIN [nt:unstructured] AS pd ON ISCHILDNODE(pd, p) 
WHERE (((LOCALNAME(p) LIKE '%12345%' OR (pd.type = 'title' AND CONTAINS(pd.content, '12345'))) 
  AND (p.[phpcr:class] = 'Path\To\Product' OR p.[phpcr:classparents] = 'Path\To\Product'))
  AND (pd.[phpcr:class] = 'Path\To\ProductDetail' OR pd.[phpcr:classparents] = 'Path\To\ProductDetail'))
```

This all works as expected, _if and only if_ the `Product` nodes contain at least one `ProductDetail` child -- in other words, it seems to behave as an `INNER JOIN`. Any `Product` node that has no such child will not be included in the results of the query, because of two things:

1. The query is specifying `SELECT *` rather than `SELECT p.*`, which is requiring a result for the `pd` alias
2. The query is appending an `AND` clause at the end, which will only match if the alias `pd` is present and of the appropriate class

The query that I actually want is this:

```sql
SELECT p.* 
FROM [nt:unstructured] AS p 
LEFT OUTER JOIN [nt:unstructured] AS pd ON ISCHILDNODE(pd, p) 
[...]

PR #667, opened 2015-11-01
Specify unique node types via mapping to allow proper outer joins

sarcher:
Because of the existence of the `class` and `classparents` query restrictions on loosely-typed nodes, it is not possible to perform real outer joins. This adds a new document-level mapping property called `uniqueNodeType` that identifies documents which can safely have those restrictions omitted, allowing for proper joins. (See discussion in #658)

The full scope of this work spans a few different repositories:

* [x] Implement mapping property in annotations, XML, and YML
* [x] Make query builder aware of new property
* [x] Update unit tests to consider new property for mapping, serialization, and deserialization
* [x] New functional test to demonstrate correctly working outer join
* [x] Create validation mechanism for unique node types
* [x] Create command for triggering validation
* [x] `phpcr-bundle` Create command for triggering validation (see doctrine/DoctrinePHPCRBundle#226)
* [x] `phpcr-bundle` Hook into cache warming for verification (same)
* [x] `phpcr-odm-documentation` Document this new feature (see doctrine/phpcr-odm-documentation#78)

Also note that this corrects a problem with `Functional\QueryBuilderTest` that was introduced in phpcr/phpcr-utils#157 and is currently failing in `master`. I could have done this in a separate PR but I needed to modify those lines for this one anyway.


sarcher:
This also requires a change to [jackalope-doctrine-dbal](https://github.com/jackalope/jackalope-doctrine-dbal), which is why the Travis build is failing currently. In `Client.php` the following:

```php
$result[] = array('dcr:name' => 'jcr:path' , 'dcr:value' => $row[$columnPrefix . 'path'], 'dcr:selectorName' => $selectorName);
$result[] = array('dcr:name' => 'jcr:score', 'dcr:value' => 0                           , 'dcr:selectorName' => $selectorName);
if (0 === count($qom->getColumns())) {
    $selectorPrefix = null !== $selector->getSelectorName() ? $selectorName . '.' : '';
    $result[] = array('dcr:name' => $selectorPrefix . 'jcr:primaryType', 'dcr:value' => $primaryType, 'dcr:selectorName' => $selectorName);
}
```

Must become:

```php
$result[] = array('dcr:name' => 'jcr:path' , 'dcr:value' => $row[$columnPrefix . 'path'], 'dcr:selectorName' => $selectorName);
$result[] = array('dcr:name' => 'jcr:score', 'dcr:value' => 0                           , 'dcr:selectorName' => $selectorName);

$selectorPrefix = null !== $selector->getSelectorName() ? $selectorName . '.' : '';
$result[] = array('dcr:name' => $selectorPrefix . 'jcr:primaryType', 'dcr:value' => $primaryType, 'dcr:selectorName' => $selectorName);
```

The reason is that to support outer joins, Jackrabbit must have the source documents specified, e.g. `SELECT a.*, b.* FROM ...`. Simply doing `SELECT * FROM...` will fail if `b` is null (perhaps this is at the transport level and not on the Jackrabbit level, but I wasn't able to isolate it; the failure seems to be in how results are delivered/converted).

To work around that, I simply made selecting each source document explicit if specific columns are not requested. So for example, if you have two document sources, it will submit a query like `SELECT a.*, b.*` rather than `SELECT *`. The implication is that the code above will never trigger, since `$qom->getColumns()` will never equal 0, and the code for hydrating rows relies on either an explicit column selector or for the `jcr:primaryType` to be present in the row result.

I've tested that particular change with my own code using both Jackrabbit and MySQL and it seems to be fine, and all tests pass. But if anybody has better guidance on how to solve that problem, I'm all ears.

sarcher:
This is now complete pending review, with all the pieces in place other than the Jackalope change mentioned above. I edited the PR description a bit so that all of the associated PRs and other info are in there.

lib/Doctrine/ODM/PHPCR/Tools/Console/Command/VerifyUniqueNodeTypesMappingCommand.php
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Doctrine\ODM\PHPCR\Tools\Console\Command;
+
+use Doctrine\ODM\PHPCR\Tools\Helper\UniqueNodeTypeHelper;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Show information about mapped documents
```
dbu: phpdoc needs to be adjusted

lib/Doctrine/ODM/PHPCR/Tools/Console/Command/VerifyUniqueNodeTypesMappingCommand.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->setName('doctrine:phpcr:mapping:verify_unique_node_types')
```
dbu: please use hyphens `-` as separator in command names

lib/Doctrine/ODM/PHPCR/Tools/Console/Command/VerifyUniqueNodeTypesMappingCommand.php
```diff
@@ -0,0 +1,60 @@
...
+            ->setDescription('Verify that documents with unique node types are correctly mapped')
```
dbu: maybe "Verify that documents claiming to have unique node types really are unique."

lib/Doctrine/ODM/PHPCR/Tools/Console/Command/VerifyUniqueNodeTypesMappingCommand.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $documentManager = $this->getHelper('phpcr')->getDocumentManager();
+
+        UniqueNodeTypeHelper::checkNodeTypeMappings($documentManager, $output);
```
dbu: we try to move away from static calls. can you please make the method non-static? its ok to instantiate the helper in the command however.

lib/Doctrine/ODM/PHPCR/Tools/Helper/UniqueNodeTypeHelper.php
```diff
@@ -0,0 +1,69 @@
...
+class UniqueNodeTypeHelper
+{
+    /**
+     * Check each mapped PHPCR-ODM document for the given document manager,
+     * throwing an exception if any document is set to use a unique node
+     * type but the node type is re-used. If an OutputInterface is provided,
+     * write some basic information to it.
+     *
+     * @param DocumentManagerInterface $documentManager The document manager to check mappings for.
+     * @param OutputInterface $output If provided, output will be written here.
```
dbu: this is leaking symfony console into the phpcr-odm domain. i would prefer to return an array with debugging information from this method. then the command can do with that whatever it wants. the array could be "FQN" => "information text". the command should then output that if we are in "debug" verbosity

dbu:
thanks, this is really well done! i will look into the jackalope issue you found. i found some minor code style issues and the like, maybe you could clean those up?

dbu:
oh well. there are indeed bugs with outer joins. started to work on them.

dbu:
see https://github.com/jackalope/jackalope-jackrabbit/pull/127 and https://github.com/jackalope/jackalope-doctrine-dbal/pull/308

is this PR strictly depending on the fix? it makes of course not much sense to only work when the outer join happens to find something in all cases, but i am not sure how long it will take us to fix jackalope-doctrine-dbal. maybe we should skip the test with jackalope-doctrine-dbal for now and merge the feature so it can make it in the next release which is imminent. as soon as dbal is fixed, we tag a new version and from there on people can use this. we can put a warning in the doc saying what the minimal required dbal version is.

dbu:
so i introduced a test for querying with * that failed, and then fix it https://github.com/phpcr/phpcr-api-tests/pull/174/files and https://github.com/jackalope/jackalope-doctrine-dbal/pull/308

can you please try to use that version of jackalope in composer.json (`dev-fix-outer-joins as 1.2.5`) to see if that solves the problem? if it does, we will merge the jackalope fixes and tag a dot release.

dbu:
also, this needs to be rebased on master, we seem to have touched something that you also change in this PR

sarcher:
I've applied the requested style fixes and rebased to the current master (the conflict was from the test I'd fixed in the PR description).

The test suite passes with the current version of Jackalope DBAL and I can't seem to duplicate any of the old problems in Jackrabbit either, so it seems to be okay for now. I will keep working with it but so far the current state here seems to be good.

sarcher:
We are seeing a related error in Travis with the DBAL now that I wasn't getting locally; that might be one for you, but if not, let me know.

dbu:
the same failure happens with master, so its not related to what you do here.

thanks a lot for this contribution!

PR #677, opened 2015-12-08
cleanup composer.json, use RMT version that supports symfony 3

dbu:


dbu:
yay, we now test with symfony 3 on travis!

PR #679, opened 2015-12-18
Fix QueryBuilder cloning issue (deep clone converter dependency)

steffenbrem:
Deep clone converter dependency on QueryBuilder, so constraints are reset when cloning a QueryBuilder instance.

Fixed issue https://github.com/Sylius/Sylius/issues/3752

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
@@ -440,10 +440,12 @@ class QueryBuilder extends AbstractNode
     }
 
     /**
-     * Deep clone on converter, so it resets it's constraints.
+     * Ensure cloned query builder objects have unique instances of the converter.
      */
     public function __clone()
     {
-        $this->converter = clone $this->converter;
+        if (null !== $this->converter) {
+            $this->converter = clone $this->converter;
+        }
     }
 }
--- a/lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
@@ -438,4 +438,12 @@ class QueryBuilder extends AbstractNode
     {
         return $this->primaryAlias;
     }
+
+    /**
+     * Deep clone on converter, so it resets it's constraints.
+     */
+    public function __clone()
+    {
+        $this->converter = clone $this->converter;
+    }
 }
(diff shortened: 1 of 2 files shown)


steffenbrem:
Is this sufficient or does the Converter needs deep cloning too?

dbu:
i am not an expert on cloning, but it seems to me this is sufficient. the converter mainly has arrays of immutable query object model objects and access to the document manager and similar things that we must not clone.

* there are two alternative scenarios: you fetch a new query builder each time, instead of cloneing
*  we rewrite the converter to work without object state, passing around the information in the methods. that sounds cleaner, but not sure how much hassle it would be

@lsmith77 is cloning a query builder a valid scenario or not recommended? unless you think cloning is totally wrong, i would merge this so it can go into the soon to come 1.3.1

lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
```diff
@@ -438,4 +438,12 @@ public function getPrimaryAlias()
     {
         return $this->primaryAlias;
     }
+
+    /**
+     * Deep clone on converter, so it resets it's constraints.
```
dantleech: This doesn't make much sense to me. Would `Ensure cloned query builder onjects have different instances of the converter` make more sense?

dantleech:
Unit test would be nice.

steffenbrem:
@dantleech I will look into that.

lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
steffenbrem: Yeah that describes it better, I will change that too :+1:

steffenbrem:
I've added a unit test. Never used `phpunit` much before, is this test okay?

tests/Doctrine/Tests/ODM/PHPCR/Query/Builder/QueryBuilderTest.php
```diff
@@ -100,4 +100,20 @@ public function testPrimaryAlias()
...
+        $property->setAccessible(true);
+
+        $this->node->setConverter($this->getMock('Doctrine\ODM\PHPCR\Query\Builder\ConverterInterface'));
+
+        $clone = clone $this->node;
+
+        $hash = spl_object_hash($property->getValue($this->node));
+        $hashClone = spl_object_hash($property->getValue($clone));
+
+        $this->assertNotEquals($hash, $hashClone, 'Cloned instance of QueryBuilder does not have an unique Converter instance.');
```
dantleech: You could simply do `$this->assertNotSame($property->getValue($this->node), $property->getValue($clone))`.

steffenbrem: Ah nice, that's a lot cleaner indeed. I have updated the unit test.

steffenbrem:
Ready.

dbu:
great job, thanks

PR #685, opened 2016-01-09
fix styleci

dbu:
we should ignore the classes that are not legal on PHP 7. they must simply not be autoloaded in PHP 7.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -3,3 +3,10 @@ preset: psr2
 enabled:
   - long_array_syntax
   - duplicate_semicolon
+
+finder:
+  not-name:
+    # exclude PHP7 illegal annotation classes to avoid false alert. those must not be attempted to be used in PHP 7
+    - Float.php
+    - String.php
+    - Int.php


.styleci.yml
```diff
@@ -2,4 +2,9 @@ preset: psr2
...
 enabled:
   - long_array_syntax
-  - duplicate_semicolon
\ No newline at end of file
+  - duplicate_semicolon
+
+finder:
+  exclude:
+    # exclude PHP7 illegal annotation classes to avoid false alert. those must not be attempted to be used in PHP 7
+    - "lib/Doctrine/ODM/PHPCR/Mapping/Annotations"
```
dbu: i tried specifying individual files here, but that did not work. also not when using not-name instead of exclude. i guess we can live with this?

lsmith77: whats the issue .. we are now ignoring a few too many files?

dbu: when i list the explicit files, i still get complaints that the files are illegal

dbu: documentation is extremely sparse, maybe there is a way to do it

GrahamCampbell: These methods exactly match the symfony finder methods.

GrahamCampbell: Exclude can be used for folders, and "not-name" to match files, as shown in our docs.

GrahamCampbell: Try without the double quotes.

.styleci.yml
```diff
@@ -2,4 +2,10 @@ preset: psr2
...
   - long_array_syntax
-  - duplicate_semicolon
\ No newline at end of file
+  - duplicate_semicolon
+
+finder:
+  not-name:
+    # exclude PHP7 illegal annotation classes to avoid false alert. those must not be attempted to be used in PHP 7
+    - lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Float.php
+    - lib/Doctrine/ODM/PHPCR/Mapping/Annotations/String.php
```
dbu: @GrahamCampbell i did this for two of the 3 classes. but i still get the error for the classes that should be ignored. can you see what is wrong with this?

GrahamCampbell: Not name strictly has to be the filename, not it's path.

GrahamCampbell: `Float.php` and `String.php` should be fine here, since you really don't want to load any classes with those names, where they're located, right?

dbu:
ah, that was it. it was not evident to me that not-name is only filenames and not paths, but this fixed it. thanks for your help @GrahamCampbell ! styleci is a really cool tool, using it in so many projects and it helps us a lot to conveniently keep clean code styles.

some constructive feedback on the documentation: i looked at the doc at https://styleci.readme.io/docs/configuration (is there any other doc i should have looked at?). that doc does not mention that things are built on the symfony finder. the not-name example of `*Stub.php` makes sense, but does not in itself make clear this could not also be a path. i would have thought Tests/**/*Mock.php might also work... it would be nice to specify the exact meaning of each config option and the supported value space. 
i am also kind of surprised by not getting an error that my not-name value was not valid - instead my value seems to have been silently ignored...

GrahamCampbell:
> it was not evident to me that not-name is only filenames and not paths

Sure, it just strictly means, "take the filename, check if it matches".

Issue #687, opened 2016-01-18
ReferenceManyCollection doesn't work with strategy 'path'


zamorem:
When using ReferenceManyCollection with "path" strategy reference nodes don't get loaded and error HTTP 403: Invalid UUID:  is reported.

It is because in ReferenceManyCollection::initialize() nodes are loaded  with SessionInterface::getNodesByIdentifier() which only accepts uuids not paths.

dbu:
do you think you can provide a pull request to make ReferenceManyCollection::initialize() check the strategy and use the correct call when its paths?

zamorem:
yes I can try that ... is it ok to check for strategy with:
$this->dm->getMetadataFactory()->getMetadataFor(get_class($this->document))->getAssociation($this->property)['strategy'] or is there some more straightforward way? I'm not that acquainted with phpcr code...

dbu:
i would prefer to add a constructor argument to the
ReferenceManyCollection. it would have to come after the locale and
could be $referenceType = 'uuid' and in UnitOfWork you pass 'path' if
the metadata says its path. you have the metadata available to pass the
property name anyways. i found 3 places of `new ReferenceManyCollection`
that would need to be updated.

it would be great if you also can add a case to the
Doctrine\Tests\ODM\PHPCR\Functional\ReferenceTest using the
RefTestObjByPath . we seem to testCreateByPath but not loading a
document that uses reference by path, hence we did not see the bug.
(maybe start with the test to see it actually fails, then see its fixed
with your fixed)

dbu:
fixed in #689

PR #689, opened 2016-02-12
Fix for #687 - loading nodes in referencemanycollection with path strategy

zamorem:
Based on suggestion in #687 I added parameter referenceType to ReferenceManyCollection 
so that nodes can be loaded either from a list of uuids or from a list of paths.

The test is in ReferenceTest::testReferenceManyPath, I had to call $referrer->references->refresh(); to invoke the failing behavior.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
+++ b/lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
@@ -31,11 +31,16 @@ use Doctrine\ODM\PHPCR\Exception\InvalidArgumentException;
  */
 class ReferenceManyCollection extends PersistentCollection
 {
+
+    const REFERENCE_TYPE_PATH = 'path';
+    const REFERENCE_TYPE_UUID = 'uuid';
+
     private $document;
     private $property;
     private $referencedNodes;
     private $targetDocument;
     private $originalReferencePaths;
+    private $referenceType;
 
     /**
      * Creates a new persistent collection.
@@ -46,8 +51,9 @@ class ReferenceManyCollection extends PersistentCollection
      * @param array           $referencedNodes An array of referenced nodes (UUID or path)
      * @param string          $targetDocument  The class name of the target documents
      * @param string          $locale          The locale to use during the loading of this collection
+     * @param string          $referenceType   Identifiers used for reference nodes in this collection, either path or default uuid
      */
-    public function __construct(DocumentManagerInterface $dm, $document, $property, array $referencedNodes, $targetDocument, $locale = null)
+    public function __construct(DocumentManagerInterface $dm, $document, $property, array $referencedNodes, $targetDocument, $locale = null, $referenceType = self::REFERENCE_TYPE_UUID)
     {
         $this->dm = $dm;
         $this->document = $document;
@@ -55,6 +61,7 @@ class ReferenceManyCollection extends PersistentCollection
         $this->referencedNodes = $referencedNodes;
         $this->targetDocument = $targetDocument;
         $this->locale = $locale;
+        $this->referenceType = $referenceType;
     }
 
     /**
@@ -96,7 +103,13 @@ class ReferenceManyCollection extends PersistentCollection
     {
         if (!$this->isInitialized()) {
             $referencedDocs = array();
-            $referencedNodes = $this->dm->getPhpcrSession()->getNodesByIdentifier($this->referencedNodes);
+            if (self::REFERENCE_TYPE_UUID === $this->referenceType) {
+                $referencedNodes = $this->dm->getPhpcrSession()->getNodesByIdentifier($this->referencedNodes);
+            } else {
+                $referencedNodes = $this->dm->getPhpcrSession()->getNodes($this->referencedNodes);
+            }
+
+
             $uow = $this->dm->getUnitOfWork();
             $uow->getPrefetchHelper()->prefetch($this->dm, $referencedNodes, $this->locale);
 
@@ -150,9 +163,13 @@ class ReferenceManyCollection extends PersistentCollection
                     $this->originalReferencePaths[] = $uow->getDocumentId($reference);
                 }
             } else {
-                $nodes = $this->dm->getPhpcrSession()->getNodesByIdentifier($this->referencedNodes);
-                foreach ($nodes as $node) {
-                    $this->originalReferencePaths[] = $node->getPath();
+                if (self::REFERENCE_TYPE_UUID === $this->referenceType) {
+                    $nodes = $this->dm->getPhpcrSession()->getNodesByIdentifier($this->referencedNodes);
+                    foreach ($nodes as $node) {
+                        $this->originalReferencePaths[] = $node->getPath();
+                    }
+                } else {
+                    $this->originalReferencePaths = $this->referencedNodes;
                 }
             }
         }
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -523,7 +523,15 @@ class UnitOfWork
                     }
 
                     $targetDocument = isset($mapping['targetDocument']) ? $mapping['targetDocument'] : null;
-                    $coll = new ReferenceManyCollection($this->dm, $document, $mapping['property'], $referencedNodes, $targetDocument, $locale);
+                    $coll = new ReferenceManyCollection(
+                        $this->dm,
+                        $document,
+                        $mapping['property'],
+                        $referencedNodes,
+                        $targetDocument,
+                        $locale,
+                        $this->getReferenceManyCollectionTypeFromMetadata($mapping)
+                    );
                     $documentState[$fieldName] = $coll;
                 }
             }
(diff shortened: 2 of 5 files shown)


lsmith77:
thanks ... can you look at the CS issues?

dbu:
great, thanks!

i have one small request: please define constants for the reference types and use those in the UoW, instead of copying the strings.

lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
```diff
@@ -96,7 +99,13 @@ public function initialize()
     {
         if (!$this->isInitialized()) {
             $referencedDocs = array();
-            $referencedNodes = $this->dm->getPhpcrSession()->getNodesByIdentifier($this->referencedNodes);
+            if ($this->referenceType == 'path') {
```
dbu: please have the constant value on the left side and use identity comparison:
`if (static::REFERENCE_TYPE_PATH === $this->referenceType) {`

lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
```diff
@@ -150,9 +159,13 @@ public function getOriginalPaths()
                     $this->originalReferencePaths[] = $uow->getDocumentId($reference);
                 }
             } else {
-                $nodes = $this->dm->getPhpcrSession()->getNodesByIdentifier($this->referencedNodes);
-                foreach ($nodes as $node) {
-                    $this->originalReferencePaths[] = $node->getPath();
+                if ($this->referenceType == 'uuid') {
```
dbu: same here. and maybe switch the cases here or above to have the same order in both.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -523,7 +523,7 @@ public function getOrCreateDocuments($className, $nodes, array &$hints = array()
                     }
 
                     $targetDocument = isset($mapping['targetDocument']) ? $mapping['targetDocument'] : null;
-                    $coll = new ReferenceManyCollection($this->dm, $document, $mapping['property'], $referencedNodes, $targetDocument, $locale);
+                    $coll = new ReferenceManyCollection($this->dm, $document, $mapping['property'], $referencedNodes, $targetDocument, $locale, $this->getReferenceManyCollectionTypeFromMetadata($mapping));
```
dbu: please change this to a multiline constructor call, like the one you add to below around line 1940.

lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
lsmith77: so as per @dbu use a constant here ..

lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
lsmith77: use a constant here

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3777,6 +3778,21 @@ private function generateUuid()
...
     }
 
+
+    /**
+     * Extracts ReferenceManyCollection type from field metadata
+     *
+     * @param $referenceFieldMetadata
+     * @return string
+     */
+    private function getReferenceManyCollectionTypeFromMetadata($referenceFieldMetadata)
```
lsmith77: use constants here ..

dbu:
you need to also adjust the xml schema: https://github.com/doctrine/phpcr-odm/blob/master/doctrine-phpcr-odm-mapping.xsd#L303-L308 - i think you need to add a new type that adds PATH to the other two types for reference many. or would it make sense that referencing a single node could also use the path strategy? might be more consistent.

can you please also start a pull request on the documentation? to my surprise, http://docs.doctrine-project.org/projects/doctrine-phpcr-odm/en/latest/reference/association-mapping.html#reference-other-documents is already claiming that path would work - is that simply wrong? or is it working for reference one but not reference many?

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3777,6 +3786,21 @@ private function generateUuid()
...
+    {
+        if (isset($referenceFieldMetadata['strategy']) && 'path' === $referenceFieldMetadata['strategy']) {
```
dbu: i think here we also want to use ReferenceManyCollection::REFERENCE_TYPE_PATH instead of 'path'

dbu:
ping @zamorem

zamorem:
Hi, sorry I've been too busy lately. I'll do my best to look at it over the weekend. Anyways the path with reference many works until you try to reload data or add anything using paths. I've stumbled upon it using sonata admin.

dbu:
so this change fixes some of the things, but not all? maybe lets wrap this up first and if you can fix reload/add, do that in a separate pull request so we can already merge this.

zamorem:
I think didn't make myself clear enough so I'll try to recap:

- the path strategy with reference many according to documentation should work, it is written in the link you pointed out
- I believe the ReferenceManyCollection works when filled by initializeFromCollection method, however it fails when reference nodes are specified in constructor by paths

What I wanted to say in the previous message was, that it might have looked like it was working when you only set the contents of the ReferenceManyCollection by collections of actual nodes, not their paths. That's why the test \Doctrine\Tests\ODM\PHPCR\Functional\ReferenceTest::testReferenceManyPath worked even with the previous version of ReferenceManyCollection until I introduced the call to refresh() method. Maybe that's why it passed unnoticed. The fix in this PR should deal with it and make it work from the Sonata admin as well.

As for the 'path' string in lib/Doctrine/ODM/PHPCR/UnitOfWork.php:3798 - this one is a different path than ReferenceManyCollection::REFERENCE_TYPE_PATH. It is I believe the 'path' from strategy metadata field, one among 'hard', 'weak' and 'path'. ReferenceManyCollection::REFERENCE_TYPE_PATH refers to internal type of ReferenceManyCollection of which there are two variants 'path' and 'uuid'. The former should probably be converted to constant as well, but I'm in favor of moving it to a different PR.

And lastly to the schema update - do you mean that 
```
<xs:simpleType name="reference-strategy">
  <xs:restriction base="xs:token">
    <xs:enumeration value="WEAK"/>
    <xs:enumeration value="HARD"/>
  </xs:restriction>
</xs:simpleType>
```
should be converted to 
```
<xs:simpleType name="reference-strategy">
  <xs:restriction base="xs:token">
    <xs:enumeration value="WEAK"/>
    <xs:enumeration value="HARD"/>
    <xs:enumeration value="PATH"/>
  </xs:restriction>
</xs:simpleType>
```
?

Is it a problem that in code 'path' strategy is spelled as lowercase 'path' and in the schema definition it is uppercase?

dbu:
okay, i think i got it now, thanks for the clarifications. yeah, converting the other "path" to a constant is even more important if its two different things. but agree that this is better done in a separate pull request.

the xml schema: afaik the upper case is converted to lower case - the current things are called WEAK and HARD in the schema, and in the code its lowercase. can you add that, i think then we are ready to merge. please also squash the commits after adding the schema update.

zamorem:
Hi, I was wondering - should I fix the CS issues, although they were introduced by rebase?

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
dbu: this one is still open

dbu:
i just reran the styleci on the master branch and it was green. no idea what happens here. we can just ignore this. if master starts showing these after merging, i will just let styleci fix master.

there is one comment that is not yet adressed - can you please handle that one as well?

do we need to update the documentation or is it all clearly documented already, and just not working as advertised?

zamorem:
Hi, I looked at the documentation and as you wrote earlier, in http://docs.doctrine-project.org/projects/doctrine-phpcr-odm/en/latest/reference/association-mapping.html#reference-other-documents it states that all strategies are valid for both ReferenceOne and ReferenceMany, which is now correct with xsd updated and this issue fixed.

The issue manifested only when dealing with ReferenceMany collection - I had no problems with ReferenceOne, which is also covered in test here: \Doctrine\Tests\ODM\PHPCR\Functional\ReferenceTest::testCreateByPath

dbu:
okay, great. so the only thing missing is https://github.com/doctrine/phpcr-odm/pull/689#discussion_r54686326 - can you adress that please?

zamorem:
Hi, concerning https://github.com/doctrine/phpcr-odm/pull/689#discussion_r54686326 I rechecked it and I think my previous comment holds:

> As for the 'path' string in lib/Doctrine/ODM/PHPCR/UnitOfWork.php:3798 - this one is a different path than ReferenceManyCollection::REFERENCE_TYPE_PATH. It is I believe the 'path' from strategy metadata field, one among 'hard', 'weak' and 'path'. ReferenceManyCollection::REFERENCE_TYPE_PATH refers to internal type of ReferenceManyCollection of which there are two variants 'path' and 'uuid'. The former should probably be converted to constant as well, but I'm in favor of moving it to a different PR.

dbu:
sorry, lost that one. thanks again @zamorem

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
dbu: https://github.com/doctrine/phpcr-odm/pull/689#issuecomment-220904994

Issue #688, opened 2016-02-04
[Bug] Cannot use string type literal value for jcr long type property


steffenbrem:
I have a document, where the `workspace` property is mapped as a `LONG`. When I execute the following query, it builds the wrong query, and returns zero rows (while I have 4 documents matching those criteria):
```php
$queryBuilder
    ->andWhere()
        ->eq()
            ->field('o.workspace')
            ->literal($value) // $value has a value of '1' (string)
;
```
This is the query statement it produces:
```
SELECT * FROM [nt:unstructured] AS o WHERE (o.[workspace] = '1' AND (o.[phpcr:class] =
'Mango\Component\Content\Model\Page' OR o.[phpcr:classparents] =
'Mango\Component\Content\Model\Page'))
```

Now, when I type cast the `$value` variable to be an int, it builds the proper query statement and casts `'1'` to `LONG`.

I think this is a bug, because you can pass in strings in the query builder for boolean type properties and those work fine.

I am using Jackrabbit 2 as my storage engine btw.

dbu:
on phpcr level, we can't know in advance what type a field is, so that would not be possible
but with the odm query builder we do know the document definition, so we should be able to do the necessary casts. this would need to happen in ConverterPhpcr - but i did not find where exactly, as it is quite generically written. if you want to work on this, ping dantleech, he might have some more specific ideas.

steffenbrem:
@dantleech ping

Thanks for pointing that out, I will try to figure out how the code is flowing in the query builder and try to come up with a solution. If @dantleech could share his thoughts on this, that would be great! :+1:

dantleech:
hi @steffenbrem 

I think a start would be to override the [walkOperandStaticLiteral](https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterBase.php#L525) method from the `ConverterBase` class in the `ConverterPhpcr` class - where you will have access to the document metadata.

Would that make sense @dbu ?

dbu:
sounds reasonable, yes. i was trying to find where i have the metadata,
did not realize the converter base class / phpcr class separation...

steffenbrem:
How would you get the document metadata from that method? I don't see how I could get it.

steffenbrem:
I've fiddled around and came up with the following, is something like this acceptable?

```php
protected function walkOperandStaticLiteral(OperandStaticLiteral $node)
{
[...]

PR #690, opened 2016-02-14
[Bug Fix] Make sure static literal value is properly cast to it's desired type

steffenbrem:
Fixes issue #688

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
@@ -340,7 +340,8 @@ class ConverterPhpcr extends ConverterBase
         if ($field = $node->getParent()->getChildOfType(AbstractLeafNode::NT_OPERAND_DYNAMIC)) {
             if ($field instanceof OperandDynamicField) {
                 $meta = $this->aliasMetadata[$field->getAlias()];
-                $type = $meta->getFieldMapping($field->getField())['type'];
+                $fieldMapping = $meta->getFieldMapping($field->getField());
+                $type = $fieldMapping['type'];
 
                 $typeMapping = array(
                     'string' => 'string',
--- a/lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
@@ -332,4 +332,35 @@ class ConverterPhpcr extends ConverterBase
 
         return $op;
     }
+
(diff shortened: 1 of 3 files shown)


lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
```diff
@@ -332,4 +332,35 @@ protected function walkOperandDynamicField(OperandDynamicField $node)
...
+
+        if ($field = $node->getParent()->getChildOfType(AbstractLeafNode::NT_OPERAND_DYNAMIC)) {
+            if ($field instanceof OperandDynamicField) {
+                $meta = $this->aliasMetadata[$field->getAlias()];
+                $type = $meta->getFieldMapping($field->getField())['type'];
+
+                $typeMapping = array(
+                    'string' => 'string',
+                    'long' => 'integer',
+                    'decimal' => 'string',
```
steffenbrem: Should this be cast to a string? According to the docs, it should be.

lsmith77:
thanks.

what issue are you trying to fix specifically?

more importantly if we are not embedding data properly into the query shouldn't this be done inside the PHPCR layer?

a failing test case would help.

steffenbrem:
Well, one use case is for example that if your query builder criteria comes from the query parameters of the `Request`, it always is a string. Since in other doctrine manager you can just pass in strings when it expects integers, I would expect the same for the `phpcr-odm`. Otherwise you would need a lot more application level logic to handle proper type casting. I think it good to eliminate this limitation at the query builder level.

For your question about implementing this in the PHPCR layer, I don't see how you can do this, since you need access to the document metadata (to fetch the fields type mapping).

lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
dbu: decimal is arbitrary large numbers, and the idea in PHPCR is to use the BCMATH extension for this. and that is based on strings.

dbu:
sounds sane to me. i would love to see a functional test that fails without your change and succeeds with it. that way we also prevent regressions.

you can ignore the styleci complaint, it is not related to your PR.

steffenbrem:
@dbu On it, I have written a test that checks that you can pass in strings for `long` fields. What do you mean by a test that fails without my change? I don't see how I can accomplish that.

steffenbrem:
I have added a functional test.

dbu:
thanks for the fix!
"test that fails without your change": if we comment out the method you added, but keep in your test, that test should fail, proving the code you provide actually fixes a real bug.

Issue #692, opened 2016-02-29
Could not find a mapped field or association named "jcr:uuid" for alias "content"


billsworld:
I am trying to use the ```queryBuilder``` to generate a query using jcr:uuid like this.

```
SELECT comments.* 
FROM [nt:unstructured] AS comments 
INNER JOIN [nt:unstructured] AS content ON content.[comments]=comments.[jcr:uuid]
```

The problem is that I am getting the error: ```Could not find a mapped field or association named "jcr:uuid" for alias "a"```

Here is the php code
```
$qb->fromDocument('AppBundle\Document\CommentDocument', 'comment')
    ->addJoinInner()
    ->right()->document('AppBundle\Document\ContentDocument', 'content')->end()
    ->condition()->equi('content.comments', 'content.uuid');
```

I have also tried to use ``->condition()->equi('content.comments', 'content.id');`` which doesn't throw the error but also does not return results.

How can I do a JOIN using jcr:uuid using the query builer?

For reference my Doctrine mapping looks like this for the id
```
    <mapped-superclass
        name="AppBundle\Document\AbstractDocument"
        >

        <mixins>
            <mixin type="mix:created" />
            <mixin type="mix:lastModified" />
        </mixins>

        <id name="id">
            <generator strategy="PARENT"/>
        </id>
...
```

dbu:
does your CommentDocument map a field you call uuid? there is a mapping
for the jcr:uuid field - you need to map that in your document, then
this should be available in the query builder

billsworld:
That's what I want to do but can't figure out how. Is this documented? Have an example?

dbu:
there is supposed to be `@Field(type=uuid)`, does that not work?

dantleech:
think we need to update the docs, as they still use the deprecated `@Uuid` but both should work: http://doctrine-orm.readthedocs.org/projects/doctrine-phpcr-odm/en/latest/reference/annotations-mapping.html#uuid²

billsworld:
Sorry. How should that look with xml mapping. Shouldn't it be used here?
https://github.com/symfony-cmf/BlockBundle/tree/master/Resources/config/doctrine-phpcr

billsworld:
It is interesting that the uuid is not used in any of the Symfony CMF code that I can find.

This seems to work for xml mapping
```
<field name="uuid" type="uuid" />
```

It is also intersting to note that this does not throw an error, there is not validation on the type
```
[...]

PR #694, opened 2016-03-01
if field type is uuid set its property to jcr:uuid

billsworld:
Fix for https://github.com/doctrine/phpcr-odm/issues/692

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
+++ b/lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
@@ -145,13 +145,17 @@ class ConverterPhpcr extends ConverterBase
 
         $propertyName = $fieldMeta['property'];
 
+        if ($fieldMeta['type'] == 'uuid') {
+            $propertyName = 'jcr:uuid';
+        }
+
         if (empty($fieldMeta['translated'])
             || empty($this->translator[$originalAlias])
         ) {
             return array($originalAlias, $propertyName);
         }
 
-        $propertyPath = $this->translator[$originalAlias]->getTranslatedPropertyPath($originalAlias, $fieldMeta['property'], $this->locale);
+        $propertyPath = $this->translator[$originalAlias]->getTranslatedPropertyPath($originalAlias, $propertyName, $this->locale);
 
         $this->aliasWithTranslatedFields[$originalAlias] = true;
 


dantleech:
Hmm, it is strange. I am pretty sure this has worked in the past, and after a quick look the mapping should happen [here](https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php#L1406). So I wonder why it isn't working in this specific case (or perhaps there has been a more general regression).

dbu:
this should indeed not be necessary. the [basic crud test](https://github.com/doctrine/phpcr-odm/blob/master/tests/Doctrine/Tests/ODM/PHPCR/Functional/BasicCrudTest.php) works without it.

@billsworld did you encounter an error without this change? any chance you can reproduce the error with a test?

dbu:
maybe something with Field(type="uuid") vs Uuid?

dbu:
looking at #692 i have a suspicion: with the annotation and just type, it would work. but with xml you need to specify the name of the field and then you would also have to specify property? can you try with <field type="uuid" name="uuid" property="jcr:uuid"/> ?

lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
```diff
@@ -145,13 +145,17 @@ protected function getPhpcrProperty($originalAlias, $odmField)
...
+        }
+
         if (empty($fieldMeta['translated'])
             || empty($this->translator[$originalAlias])
         ) {
             return array($originalAlias, $propertyName);
         }
 
-        $propertyPath = $this->translator[$originalAlias]->getTranslatedPropertyPath($originalAlias, $fieldMeta['property'], $this->locale);
+        $propertyPath = $this->translator[$originalAlias]->getTranslatedPropertyPath($originalAlias, $propertyName, $this->locale);
```
dbu: this change makes sense regardless of the rest, its weird that we extract the value into a variable but then only use it once.

dbu:
reading ClassMetadata::validateAndCompleteFieldMapping it feels like uuid should work without explicitly specifying property, though. the point dan linked to hardcodes "property" to jcr:uuid, and validateAndCompleteFieldMapping only sets property if its not already set. and yes, hardcoding to jcr:uuid comes before validating.

lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
```diff
@@ -145,13 +145,17 @@ protected function getPhpcrProperty($originalAlias, $odmField)
 
         $propertyName = $fieldMeta['property'];
 
+        if ($fieldMeta['type'] == 'uuid') {
```
dbu: https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php#L1407 should lead to this condition never being true

billsworld:
```field type="uuid" name="uuid" property="jcr:uuid"/>``` does the trick. There was even validation if the document is not referenceable.

PR #696, opened 2016-03-02
reuse cached $fieldMeta['property'] variable

billsworld:
the good part from https://github.com/doctrine/phpcr-odm/pull/694


dbu:
thanks!

PR #702, opened 2016-04-18
[WIP] Add mapping to index reference collection by node name

### LINKED ISSUE #701: Support for "Indexed Associations" as in the Doctrine ORM
I could not find a way to make a `reference-many` association indexed by a specific field or it's node name. In the Doctrine ORM there is something like the `indexBy` mapping, that you can use to get an associative collection. Is support for this in PHPCR ODM something that might happen?

We could achieve this by adding the appropriate code to the `ReferenceManyCollection::initialize` method.

koemeet:
Fixes #701 
- [ ] Implement annotations and YAML driver
- [ ] Restrict it only to only `reference-many` collections?

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
+++ b/lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
@@ -100,13 +100,22 @@ class ReferenceManyCollection extends PersistentCollection
             $uow = $this->dm->getUnitOfWork();
             $uow->getPrefetchHelper()->prefetch($this->dm, $referencedNodes, $this->locale);
 
+            $metadata = $this->dm->getClassMetadata(get_class($this->document));
+            $metadata = $metadata->mappings[$this->property];
+            $isAssociative = $metadata['index-by-nodename'];
+
             $this->originalReferencePaths = array();
             foreach ($referencedNodes as $referencedNode) {
                 $proxy = $uow->getOrCreateProxyFromNode($referencedNode, $this->locale);
                 if (isset($targetDocument) && !$proxy instanceof $this->targetDocument) {
                     throw new PHPCRException("Unexpected class for referenced document at '{$referencedNode->getPath()}'. Expected '{$this->targetDocument}' but got '".ClassUtils::getClass($proxy)."'.");
                 }
-                $referencedDocs[] = $proxy;
+                if ($isAssociative) {
+                    $index = $referencedNode->getName();
+                    $referencedDocs[$index] = $proxy;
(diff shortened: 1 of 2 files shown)


steffenbrem:
@dbu  So, for a `reference-many` collection it should be okay to maybe do this behaviour by default? Then we could even get rid of the whole `index-by-nodename` mapping.

dbu:
for the ReferenceManyCollection, i am not sure if we can do the associative by name at all, because of the risk of name collisions. maybe we should use the full path instead of the name for the reference many collection.
with references, you can reference arbitrary documents. if you reference /a/b/c and /b/b/c then both local names will be "c".

children are handled with the [ChildrenCollection](https://github.com/steffenbrem/phpcr-odm/blob/feature/indexed-associations/lib/Doctrine/ODM/PHPCR/ChildrenCollection.php). for children collection, i agree that having them associative would be a good idea. 

i think we should have the configuration option for now, but tell that in version 2 we will only have associative children and remove the option again. there is a risk of unexpected behaviour when the keys change.

lib/Doctrine/ODM/PHPCR/ReferenceManyCollection.php
```diff
@@ -100,13 +100,22 @@ public function initialize()
...
             $this->originalReferencePaths = array();
             foreach ($referencedNodes as $referencedNode) {
                 $proxy = $uow->getOrCreateProxyFromNode($referencedNode, $this->locale);
                 if (isset($targetDocument) && !$proxy instanceof $this->targetDocument) {
                     throw new PHPCRException("Unexpected class for referenced document at '{$referencedNode->getPath()}'. Expected '{$this->targetDocument}' but got '".ClassUtils::getClass($proxy)."'.");
                 }
-                $referencedDocs[] = $proxy;
+                if ($isAssociative) {
+                    $index = $referencedNode->getName();
+                    $referencedDocs[$index] = $proxy;
```
dbu: this might overwrite other referenced documents with the same node name. we should use $referenceNode->getPath() i think.

dbu:
i just started with a 2.0 branch. maybe this should be done against 2.0 so we need not worry about unexpected BC troubles because of slightly different behaviour. wdyt @steffenbrem ?

dbu:
@steffenbrem do you want to wrap this up for 2.0? 

i think we need to index by path instead of by nodename, to avoid problems with name collisions.

steffenbrem:
@dbu Nice! I will take a look at it and implement it to index by path, seems indeed a better approach.

PR #706, opened 2016-05-30
Support child restriction

dantleech:
**TODO**
* [x] Wrap up feature
* [x] Documentation (also mention that if a document already is there because the situation was created before the annotation was used or from raw phpcr operations, the child will be on the document as validation only happens during save but not on read)

Doc PR: https://github.com/doctrine/phpcr-odm-documentation/pull/81

This PR allows the restriction of child document types:

```php
/**
 * @PHPCRODM\Document(restrictChildren={
 *   "Doctrine\Tests\Models\CMS\CmsArticle"
 * })
 */
 class ArticleFolder
 {
     // ...
 }
```

- By default children are unrestricted (`childRestrictions = null`)
- Children validated on "insert".
- Children validated on "move".

### Why?

- Ensure structural integrity where it matters.
- Determine which types can be added to a document (e.g. in a tree browser).
- Ultimately this could also be part of the node type definition if we ever implementing that.

### Alternatives

We could also implement this as a separate annotation:

```php
/**
 * @Document()
 * @RestrictChildren({
 *   "AppBundle\Document\Foo"
 *   "AppBundle\Document\Bar"
 * }, restrict="include")
 */
```

 Where ``restrict`` is an enum of `none`, `include` and `exclude`.

### DIFF
--- a/doctrine-phpcr-odm-mapping.xsd
+++ b/doctrine-phpcr-odm-mapping.xsd
@@ -44,6 +44,10 @@
         <xs:anyAttribute namespace="##other"/>
     </xs:complexType>
 
+    <xs:complexType name="child-class">
+        <xs:attribute name="name" type="xs:string"/>
+    </xs:complexType>
+
     <xs:complexType name="mixin">
         <xs:attribute name="type" type="xs:NMTOKEN"/>
     </xs:complexType>
@@ -138,6 +142,7 @@
             <xs:element name="document-listeners" type="phpcr:document-listeners" minOccurs="0" maxOccurs="1" />
             -->
             <xs:element name="id" type="phpcr:id" minOccurs="0" maxOccurs="1" />
+            <xs:element name="child-class" type="phpcr:child-class" minOccurs="0" maxOccurs="unbounded" />
             <xs:element name="uuid" type="phpcr:uuid" minOccurs="0" maxOccurs="1" />
             <xs:element name="locale" type="phpcr:locale" minOccurs="0" maxOccurs="1" />
             <xs:element name="field" type="phpcr:field" minOccurs="0" maxOccurs="unbounded"/>
@@ -159,6 +164,7 @@
         <xs:attribute name="translator" type="xs:NMTOKEN"/>
         <xs:attribute name="versionable" type="phpcr:versionable-type" />
         <xs:attribute name="referenceable" type="xs:boolean" default="false" />
+        <xs:attribute name="is-leaf" type="xs:boolean" default="false" />
         <xs:attribute name="node-type" type="xs:NMTOKEN" />
         <xs:attribute name="repository-class" type="xs:string"/>
         <xs:anyAttribute namespace="##other"/>
--- a/lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Document.php
+++ b/lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Document.php
@@ -66,4 +66,14 @@ class Document
      * @var boolean
      */
     public $uniqueNodeType;
+
+    /**
+     * @var array
+     */
+    public $childClasses = array();
+
+    /**
+     * @var boolean
+     */
+    public $isLeaf;
 }
--- a/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
@@ -31,6 +31,7 @@ use Doctrine\Common\Persistence\Mapping\ClassMetadata as ClassMetadataInterface;
 use Doctrine\Common\ClassLoader;
 use Doctrine\Instantiator\Instantiator;
 use Doctrine\Instantiator\InstantiatorInterface;
+use Doctrine\ODM\PHPCR\Exception\OutOfBoundsException;
 
 /**
  * Metadata class
@@ -329,6 +330,23 @@ class ClassMetadata implements ClassMetadataInterface
     public $parentClasses = array();
 
     /**
+     * READ-ONLY: Child class restrictions.
+     *
+     * If empty then any classes are permitted.
+     *
+     * @var array
+     */
+    public $childClasses = array();
+
+    /**
+     * READ-ONLY: If the document should be act as a leaf-node and therefore
+     *            not be allowed children.
+     *
+     * @var boolean
+     */
+    public $isLeaf = false;
+
+    /**
      * The inherited fields of this class
      *
      * @var array
@@ -451,6 +469,57 @@ class ClassMetadata implements ClassMetadataInterface
                     break;
             }
         }
+    }
+
+    /**
+     * Validate that childClasses is empty if isLeaf is true.
+     *
+     * @throws MappingException if there is a conflict between isLeaf and childClasses.
+     */
+    public function validateChildClasses()
+    {
+        if (count($this->childClasses) > 0 && $this->isLeaf) {
+            throw new MappingException(sprintf(
+                'Cannot map a document as a leaf and define child classes for "%s"',
+                $this->name
+            ));
+        }
+    }
+
+    /**
+     * Assert that the given class FQN can be a child of the document this
+     * metadata represents.
+     *
+     * @param string $classFqn
+     * @throws OutOfBoundsException
+     */
+    public function assertValidChildClass(ClassMetadata $class)
+    {
+        if ($this->isLeaf()) {
+            throw new OutOfBoundsException(sprintf(
+                'Document "%s" has been mapped as a leaf. It cannot have children',
+                $this->name
+            ));
+        }
+
+        $childClasses = $this->getChildClasses();
+
+        if (0 === count($childClasses)) {
+            return;
+        }
+
+        foreach ($childClasses as $childClass) {
+            if ($class->name === $childClass || $class->reflClass->isSubclassOf($childClass)) {
+                return;
+            }
+        }
+
+        throw new OutOfBoundsException(sprintf(
+            'Document "%s" does not allow children of type "%s". Allowed child classes "%s"',
+            $this->name,
+            $class->name,
+            implode('", "', $childClasses)
+        ));
     }
 
     /**
@@ -1122,6 +1191,49 @@ class ClassMetadata implements ClassMetadataInterface
     {
         return $this->parentClasses;
     }
+
+    /**
+     * Return the class names or interfaces that children of this document must
+     * be an instance of.
+     *
+     * @return string[]
+     */
+    public function getChildClasses()
+    {
+        return $this->childClasses;
+    }
+
+    /**
+     * Set the class names or interfaces that children of this document must be
+     * instance of.
+     *
+     * @param string[] $childClasses
+     */
+    public function setChildClasses(array $childClasses)
+    {
+        $this->childClasses = $childClasses;
+    }
+
+    /**
+     * Return true if this is designated as a leaf node.
+     *
+     * @return bool
+     */
+    public function isLeaf()
+    {
+        return $this->isLeaf;
+    }
+
+    /**
+     * Set if this document should act as a leaf node.
+     *
+     * @param bool $isLeaf
+     */
+    public function setIsLeaf($isLeaf)
+    {
+        $this->isLeaf = $isLeaf;
+    }
+    
 
     /**
      * Checks whether the class will generate an id via the repository.
--- a/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadataFactory.php
+++ b/lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadataFactory.php
@@ -256,6 +256,7 @@ class ClassMetadataFactory extends AbstractClassMetadataFactory
         $class->validateIdentifier();
         $class->validateReferenceable();
         $class->validateReferences();
+        $class->validateChildClasses();
         $class->validateLifecycleCallbacks($this->getReflectionService());
         $class->validateTranslatables();
 
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -54,6 +54,7 @@ use PHPCR\Util\UUIDHelper;
 use PHPCR\Util\PathHelper;
 use PHPCR\Util\NodeHelper;
 use Jackalope\Session as JackalopeSession;
+use Doctrine\ODM\PHPCR\Exception\OutOfBoundsException;
 
 /**
  * Unit of work class
@@ -2332,6 +2333,8 @@ class UnitOfWork
             }
 
             $parentNode = $this->session->getNode(PathHelper::getParentPath($id));
+            $this->validateChildClass($parentNode, $class);
+
             $nodename = PathHelper::getNodeName($id);
             $node = $parentNode->addNode($nodename, $class->nodeType);
             if ($class->node) {
@@ -2730,6 +2733,9 @@ class UnitOfWork
                     $invoke
                 );
             }
+
+            $parentNode = $this->session->getNode(PathHelper::getParentPath($targetPath));
+            $this->validateChildClass($parentNode, $class);
 
             $this->session->move($sourcePath, $targetPath);
 
@@ -3967,5 +3973,24 @@ class UnitOfWork
         if ($this->eventManager->hasListeners($eventName)) {
             $this->eventManager->dispatchEvent($eventName, $event);
         }
+    }
+
+    /**
+     * If the parent node has child restrictions, ensure that the given
+     * class name is within them.
+     *
+     * @param NodeInterface $parentNode
+     * @param string $classFqn
+     */
+    private function validateChildClass(NodeInterface $parentNode, ClassMetadata $class)
+    {
+        $parentClass = $this->documentClassMapper->getClassName($this->dm, $parentNode);
+
+        if (null === $parentClass) {
+            return;
+        }
+
+        $metadata = $this->dm->getClassMetadata($parentClass);
+        $metadata->assertValidChildClass($class);
     }
 }
--- a/tests/Doctrine/Tests/Models/CMS/CmsBlogInvalidChild.php
+++ b/tests/Doctrine/Tests/Models/CMS/CmsBlogInvalidChild.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Documents;
+
+namespace Doctrine\Tests\Models\CMS;
+
+use Doctrine\ODM\PHPCR\Mapping\Annotations as PHPCRODM;
+use Doctrine\ODM\PHPCR\DocumentRepository;
+use Doctrine\ODM\PHPCR\Id\RepositoryIdInterface;
+
+/**
+ * @PHPCRODM\Document()
+ */
+class CmsBlogInvalidChild
+{
+    /** @PHPCRODM\Id(strategy="parent") */
+    public $id;
+
+    /** @PHPCRODM\NodeName() */
+    public $name;
+
+    /** @PHPCRODM\ParentDocument() */
+    public $parent;
+}
--- a/tests/Doctrine/Tests/ODM/PHPCR/Functional/UnitOfWorkTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Functional/UnitOfWorkTest.php
@@ -9,6 +9,13 @@ use Doctrine\Tests\Models\CMS\CmsAddress;
 use Doctrine\Tests\Models\References\ParentNoNodeNameTestObj;
 use Doctrine\Tests\Models\References\ParentTestObj;
 use Doctrine\Tests\Models\Translation\Comment;
+use Doctrine\Tests\Models\CMS\CmsArticleFolder;
+use Doctrine\Tests\Models\CMS\CmsGroup;
+use Doctrine\Tests\Models\CMS\CmsArticle;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Tests\Models\CMS\CmsBlogPost;
+use Doctrine\Tests\Models\CMS\CmsBlogInvalidChild;
+use Doctrine\Tests\Models\CMS\CmsBlogFolder;
 
 /**
  * @group functional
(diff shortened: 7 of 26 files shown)


lib/Doctrine/ODM/PHPCR/DocumentClassMapper.php
```diff
@@ -50,8 +52,8 @@ public function getClassName(DocumentManagerInterface $dm, NodeInterface $node,
     {
         $className = $this->expandClassName($dm, $className);
 
-        if ($node->hasProperty('phpcr:class')) {
-            $nodeClassName = $node->getProperty('phpcr:class')->getString();
+        if ($node->hasProperty(self::CLASS_PROPERTY)) {
+            $nodeClassName = $node->getProperty(self::CLASS_PROPERTY)->getString();
```
dantleech: Grep to see if `phpcr:class` is referenced anywhere else.

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Document.php
```diff
@@ -66,4 +66,9 @@ class Document
      * @var boolean
      */
     public $uniqueNodeType;
+
+    /**
+     * @var array
```
dantleech: `array|null`.

The other option is to have this as a separate annotation, `@RestrictChidren({"Foo\ChildClass", restrict="whitelist")`. Which would also help us here as restrict woule be an `enum`: `none, whitelist, blacklist` or similar.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3968,4 +3974,43 @@ public function invokeGlobalEvent($eventName, EventArgs $event)
...
+        $metadata = $this->dm->getClassMetadata($parentClass);
+        $childRestrictions = $metadata->getChildRestrictions();
+
+        if (null === $childRestrictions) {
+            return;
+        }
+
+        if ([] === $childRestrictions) {
+            throw new OutOfBoundsException(sprintf(
+                'Document "%s" cannot have children',
```
dantleech: Test

dantleech:
Would need to add XML and YAML drivers.

dbu:
interesting proposal! indeed the outside tree definition we currently have is awkward and weak. and i always forget to add my new classes to it.

i wonder if we should do this on the class level or on Child / Children mappings and then merge all Child / Children mappings to know the allowed children. or maybe keep what you propose, but add that when the annotation / attribute is present the child and children types are merged into the allowed list automatically. then i can potentially limit my model by having children explicitly mapped and just set an empty annotation / attribute to trigger restriction. i guess restricting by default is not that user friendly (and a huge BC break that can be annoying even when upgrading to 2.0 at some point)

i think i prefer the separate annotation, given that we could have options. it could be something like
```
/**
 * @Document()
 * @RestrictChildren(
 *    include={"AppBundle\Model\MarkerInterface"}, 
 *    exclude={"AppBundle\Document\SpecialCase"}
 * )
 */
```

dantleech:
> i wonder if we should do this on the class level or on Child / Children mappings and then merge all Child / Children mappings to know the allowed children

I did think about doing it on the `Children` mapping - but that applies to a property (i.e. there are many properties even if you can only map one as having the Children) - you might not even want to map the children to a property.

I think its correct to say "this document can have these children" rather than "this document which has a property mapped as children wiith these types can have those children".

> i guess restricting by default is not that user friendly (and a huge BC break that can be annoying even when upgrading to 2.0 at some point)

If we add an extra annotation then I suggest that the Metadata has an enum `$childRestriction` with the values `none`, `inclusive` and `exclusive`, with the default being `none`.

So the annotation would be fully expressed:

```
 * @RestrictChildren(
 *    classes={"AppBundle\Model\MarkerInterface"}, 
 *    type="include"
 * )
```

If `classes` are present then we could automatically switch from `none` to `include`.

dbu:
i prefer to have the explicit lists as it allows you to allow everything of a base class but exclude some exceptions. like the sonata admin extensions do. an empty `@RestrictChildren` would say there can be no children.

and i would merge types from Child / Children mappings into the allowed children. otherwise you need to duplicate that. there can be several Children annotations, btw. you can restrict them to prefixes and use a naming pattern to separate children. avoids you to have Child with a container object.

dantleech:
> I prefer to have the explicit lists (see the example i created) as it allows you to allow everything of a base class but exclude some exceptions.

ok thats makes sense.

> and i would merge types from Child / Children mappings into the allowed children. otherwise you need to duplicate that.

I don't think you can set type for Child / Children? only a node-name filter?

dbu:
> I don't think you can set type for Child / Children? only a node-name
> filter?

ups, you are right. i guess the feature would make sense, but we don't
have to mix it into this PR. maybe we could have the minimal sanity
check that when you forbid all children but have a Child / Children
mapping, we raise an error that this makes no sense?

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -329,6 +329,13 @@ class ClassMetadata implements ClassMetadataInterface
     public $parentClasses = array();
 
     /**
+     * READ-ONLY: Child class restrictions.
+     *
+     * @var array
+     */
+    public $childRestrictions = null;
```
dbu: child or children? not sure which is the correct way. its the restriction on each child...

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -2731,6 +2734,9 @@ private function executeMoves($documents)
                 );
             }
 
+            $parentNode = $this->session->getNode(PathHelper::getParentPath($targetPath));
```
dbu: this comes at a cost of loading another document

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3968,4 +3974,43 @@ public function invokeGlobalEvent($eventName, EventArgs $event)
...
+        }
+
+        if (array() === $childRestrictions) {
+            throw new OutOfBoundsException(sprintf(
+                'Document "%s" cannot have children',
+                $parentClass
+            ));
+        }
+
+        if (!in_array($classFqn, $childRestrictions)) {
```
dbu: i think you need to loop and instanceof to allow base classes or interfaces in the list.

i guess specifying a namespace instead of a class makes little sense? and we could not handle instanceof in that scenario.

tests/Doctrine/Tests/ODM/PHPCR/Functional/UnitOfWorkTest.php
```diff
@@ -185,4 +188,58 @@ public function testFetchingMultipleHierarchicalObjectsWithChildIdFirst()
...
+    {
+        $articleFolder = new CmsArticleFolder();
+        $articleFolder->id = '/functional/articles';
+
+        $article = new CmsGroup();
+        $article->id = '/functional/articles/address';
+        $article->name = 'invalid-child';
+
+        $this->dm->persist($articleFolder);
+        $this->dm->persist($article);
```
dbu: can you please add another test where the child would be persisted through cascading because its mapped on the parent document, instead of explicitly persist? i guess that also works, but we should make sure.

dbu: also, can you please add a move test where we try to move a child that is not allowed under a parent?

dbu:
what if i *edit* a document that is in an illegal place already, but is neither moved nor created? i think it would make sense to allow that - a user otherwise could have a hard time saving their changes and understanding what is happening. and the situation was illegal before the edit already...

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
dantleech: .. or rather a node. in some cases that node will already have been loaded (f.e. parent mappings *always* load the parent node), I will benchmark to check the performance cost.

dantleech:
> what if i edit a document that is in an illegal place already

Not sure .. if their tree is in an illegal state, then its broken, so I would expect an \Exception when editing an illegal child - otherwise nobody knows that it is broken until part of the system that relies on chldren being of a specific type recieves a class of the wrong type and throws an error at a critical moment.

But it does raise the issue of migration when restrictions are added to an existing model - but I would say that that excerise could be left to the user with the `PhpcrMigrationsBundle.`

dbu:
what is the current code doing?
i guess we definitely don't want to throw errors on loading. but on
saving an edit i don't know which is better, error or not. i tend to say
no error if its just property changes, but maybe its too difficult to
separate from adding another child or a child to the illegal child or
whatever, so an error sounds ok.

we could do a command to validate the validity of the current tree, to
help figure out if the migration worked as expected...

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
dantleech: if we were to follow from PHPCR we could just have `requiredClasses` as with `NodeDefinition::getRequiredPrimaryTypes`

dantleech: Also interesting that they do not permit blacklisting (afaict)

dantleech: But maybe this is an argument to keep it as a whitelist - as if we do allow mapping by phpcr_type in the future, this feature would not map to it.

dantleech:
> we could do a command to validate the validity of the current tree, to
help figure out if the migration worked as expected...

yeah that crossed my mind too.. I would say that could be another PR.

dbu:
>     we could do a command to validate the validity of the current tree, to
>     help figure out if the migration worked as expected...
> 
> yeah that crossed my mind too.. I would say that could be another PR.

yes, for sure. there are other things to check, like missing required
properties.

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
dbu: actually yes, lets only use whitelisting then. finer grained stuff can
be done with listeners or custom validators and whatnot. and go with
requiredClasses - i like to keep this as close to PHPCR as makes sense.

but i think PHPCR also does the equivalent of an instanceof check, so we
want to keep doing that.

dantleech: Shall we go with a `requiredClasses` attribute or a separate annotation?

lib/Doctrine/ODM/PHPCR/DocumentClassMapper.php
dantleech: Actually this is too risky (adding class constants to places that might be untested) and unrequired for this PR, reverting.

tests/Doctrine/Tests/ODM/PHPCR/Functional/UnitOfWorkTest.php
dantleech: Done.

dantleech:
Made the benchmark, no noticable difference:

```
+------------------+--------------------------------------+------------------------+
| subject          | vcs_branch:children_restriction:mean | vcs_branch:master:mean |
+------------------+--------------------------------------+------------------------+
| benchPersistTwo  | 40.43ms                              | 39.67ms                |
| benchPersistMany | 65.65ms                              | 72.44ms                |
+------------------+--------------------------------------+------------------------+
```

lib/Doctrine/ODM/PHPCR/DocumentClassMapper.php
dbu: agreed. but its a good idea to do, maybe in a separate PR once this one
is merged?

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
dbu: attribute is better imho, now that we decided it can be simply a list of
allowed classes

dbu:
i don't like `@Document(requiredClasses={...})`. it feels like we would define the classes *this* document may have, which would make no sense but could be confusing. i think `allowedChildClasses` would be better.

in [jackrabbit cnd](http://jackrabbit.apache.org/jcr/node-type-notation.html) the child types are defined on the child / children names. i think we have the same here. on an `@Child/Children` annotation, we don't need to repeat its about the children, but on the class level we should. requiredChildClasses feels like the children has to implement all of the classes, or we require children, but what we do is only allow children that are of one of the specified classes.

dbu:
and thanks for the benchmark, looks like its not a noticable performance decrease. and its for write only, not for read so it should be pretty safe.

dbu:
regarding forbidding all children: what if we allow to just have one entry with `false` and interpret that as not allowing children? in xml, it would be `<allowed-child-class>false</...>` which feels ok-ish. if you don't like it, a separate option would be ok. they would need to be sanity checked as the two things are mutually exclusive.

dantleech:
>> and go with requiredClasses - i like to keep this as close to PHPCR as makes sense.

> i don't like @Document(requiredClasses={...})

:) but also fine to change it, its just for the sake of copying PHPCR.

dantleech:
@dbu updated.

Have opted for the following (which I am still happy to change):

- `childClasses`: `requiredChildClasses` seemed too verbose.
- `isLeaf`: Instead of `allowChildren` as it seemed better tree terminology.

There is validation to ensure that `isLeaf` and `childClasses` do not conflict.

Will start on a doc PR.

dantleech:
Documentation added.

doctrine-phpcr-odm-mapping.xsd
```diff
@@ -44,6 +44,15 @@
         <xs:anyAttribute namespace="##other"/>
     </xs:complexType>
 
+    <xs:complexType name="child-class">
+        <xs:simpleContent>
+            <xs:extension base="xs:string">
+                <xs:attribute name="value" type="xs:string">
+                </xs:attribute>
```
dbu: should be self-closing tag

dbu: @WouterJ what is best practice here? using an element with an attribute or make the class the text content of the xml tag?

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -1124,6 +1151,47 @@ public function getParentClasses()
     }
 
     /**
+     * Return the classes that are required by children of this document
```
dbu: Return the class names or interfaces that children of this document must be instance of.

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -1124,6 +1151,47 @@ public function getParentClasses()
...
+     *
+     * @return string[]
+     */
+    public function getChildClasses()
+    {
+        return $this->childClasses;
+    }
+
+    /**
+     * Set the classes that are required by children of this document
```
dbu: Set the class names or interfaces that children of this document must be instance of.

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -1124,6 +1151,47 @@ public function getParentClasses()
...
+    {
+        $this->childClasses = $childClasses;
+    }
+
+    /**
+     * Return true if this is designated as a leaf node.
+     *
+     * @return bool
+     */
+    public function getIsLeaf()
```
dbu: would this not be `isLeaf`?

lib/Doctrine/ODM/PHPCR/Tools/Console/Command/DocumentMigrateClassCommand.php
```diff
@@ -79,7 +79,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $mapper = $documentManager->getConfiguration()->getDocumentClassMapper();
 
         $input->setOption('query', sprintf(
-            'SELECT * FROM [nt:base] WHERE [phpcr:class] = "%s"',
+            'SELECT * FROM [nt:base] WHERE [%s] = "%s"',
```
dbu: revert %s to phpcr:class please

lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Document.php
```diff
@@ -66,4 +66,19 @@ class Document
      * @var boolean
      */
     public $uniqueNodeType;
+
+    /**
+     * @var boolean
+     */
+    public $allowChildren = 'any';
```
dbu: this seems unused

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -329,6 +329,20 @@ class ClassMetadata implements ClassMetadataInterface
     public $parentClasses = array();
 
     /**
+     * READ-ONLY: Child class restrictions.
```
dbu: the phpdoc should explain that if this list is empty, it means we don't check the classes of the children. to make a leaf node, use the explicit isLeaf field.

dbu:
i like both naming choices! just some minor details that i commented on.

dbu:
please add a note to the CHANGELOG.md as well.

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
dantleech: I guess that is what would be expected.

doctrine-phpcr-odm-mapping.xsd
```diff
@@ -44,6 +44,10 @@
         <xs:anyAttribute namespace="##other"/>
     </xs:complexType>
 
+    <xs:complexType name="child-class">
+        <xs:attribute name="name" type="xs:string"/>
```
dantleech: @dbu changed to an attribute as there are no examples of node-value elements in Doctrine. /cc @WouterJ

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
dantleech: Done.

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -454,6 +471,19 @@ public function validateIdentifier()
     }
 
     /**
+     * Validate that childClasses is empty if isLeaf is true.
+     */
```
dantleech: `@throws`

tests/Doctrine/Tests/Models/CMS/CmsBlogInvalidChild.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
```
dantleech: We should add the header to styleci.

dantleech:
Updated.

dantleech:
@dbu added changelog. Build was failing due to PHPUnit warnings about mocks on two of the jobs (and only on the PR build).

dantleech:
ping @dbu

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadataFactory.php
```diff
@@ -256,6 +256,7 @@ protected function validateRuntimeMetadata($class, $parent)
         $class->validateIdentifier();
         $class->validateReferenceable();
         $class->validateReferences();
+        $class->validateChildClasses();
```
dbu: kind of strange that we don't have just one validate method on ClassMetadata and call that here, and keep the details of what needs to be validated local to ClassMetadata...

but not subject of this PR, so lets keep it like this here.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3968,4 +3974,43 @@ public function invokeGlobalEvent($eventName, EventArgs $event)
...
+                'Document "%s" has been mapped as a leaf. It cannot have children',
+                $parentClass
+            ));
+        }
+
+        if (array() === $childClasses) {
+            return;
+        }
+
+        if (!in_array($classFqn, $childClasses)) {
```
dbu: shouldn't this do instanceof checks of $classFqn against each $childClasses? as is, interfaces and inheritance in general won't work.

tests/Doctrine/Tests/Models/CMS/CmsBlogInvalidChild.php
dbu: i think styleci ignores the test folder.

dbu:
https://github.com/doctrine/phpcr-odm/pull/706#discussion_r66420203 seems a relevant question to me. i thought i asked already, maybe github decided to hide my comment... apart from that, looks really good and nice to me!

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
dantleech: indeed, well caught.. at least for subclass.

not sure if it makes sense for interfaces though -- we have no analogy for that in PHPCR - we have mixins ≈ traits and classes = node types, but interfaces?

maybe they could also be modelled by mixins?

dbu: on the php level, the interfaces do make a lot of sense, so i think we
should support them. anyways afaik instanceof does not distinguish
between interface and class. so lets not explicitly prevent interfaces here.

dantleech: I suppose we could support, is, if an interface is specified, automatically expand the PHPCR node type restriction to include all mapped classes implementing that interface. That would work wouldn't it?

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -454,6 +472,63 @@ public function validateIdentifier()
...
+    }
+
+    /**
+     * Assert that the given class FQN can be a child of the document this
+     * metadata represents.
+     *
+     * @param string $classFqn
+     * @throws OutOfBoundsException
+     */
+    public function assertValidChildClass(ClassMetadata $class)
```
dantleech: @dbu moved the validation logic to the `ClassMetadata`.

tests/Doctrine/Tests/ODM/PHPCR/Mapping/ClassMetadataTest.php
```diff
@@ -324,6 +324,91 @@ public function testClassMetadataInstanceSerializationTranslationProperties($cm)
...
     }
+
+    /**
+     * It should throw an exception if given a child class FQN when the
+     * metadata is for a leaf.
+     *
+     * @expectedException \Doctrine\ODM\PHPCR\Exception\OutOfBoundsException
+     * @expectedExceptionMessage has been mapped as a leaf
+     */
+    public function testAssertValidChildClassesIsLeaf()
```
dantleech: Much easier to test the classmetadata rather than the UOW for all these code paths.

dantleech:
@dbu updated. as indicated above I guess we can handle interfaces sensibly in PHPCR when the time comes, so now accepting classes + interfaces.

dantleech:
tests failing locally, but they normally do for random reasons, waiting to see what travis says.

dantleech:
Failures are from warnings as normal.

WouterJ:
@dantleech add `<env name="SYMFONY_DEPRECATIONS_HELPER" value="disabled"/>` to the `<php>` section of `phpunit.xml.dist` and you won't see the warnings.

tests/Doctrine/Tests/ODM/PHPCR/Mapping/ClassMetadataTest.php
```diff
@@ -324,6 +324,91 @@ public function testClassMetadataInstanceSerializationTranslationProperties($cm)
...
+     */
+    public function testAssertValidChildClassesNotAllowed()
+    {
+        $cm = new ClassMetadata('Doctrine\Tests\ODM\PHPCR\Mapping\Person');
+        $cm->initializeReflection(new RuntimeReflectionService());
+        $childCm = new ClassMetadata('stdClass');
+        $childCm->initializeReflection(new RuntimeReflectionService());
+        $cm->setChildClasses(array('Doctrine\Tests\ODM\PHPCR\Mapping\Person'));
+        $result = $cm->assertValidChildClass($childCm);
+        $this->assertNull($result);
```
dbu: this line can be removed, right?

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -454,6 +472,63 @@ public function validateIdentifier()
...
+    {
+        $childClasses = $this->getChildClasses();
```
dbu: move below the isLeaf check?

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -454,6 +472,63 @@ public function validateIdentifier()
...
+        }
+
+        if (array() === $childClasses) {
+            return;
+        }
+
+        $found = false;
+        foreach ($childClasses as $childClass) {
+            if ($class->name === $childClass || $class->reflClass->isSubclassOf($childClass)) {
+                $found = true;
```
dbu: you can simply return here and abort the loop. and get rid of the $found variable, as you do nothing additional with it.

dantleech:
Updated.

dbu:
looks good. i guess we should bump the branch-alias of master to 1.4, as this is a new feature. can you do that and start the doc PR? apart from that looks ready to merge to me.

dbu:
created #709 for the build system fix. i want the 1.3 branch to also not fail... once we merged that, please rebase this branch on master to solve the conflict on the phpunit xml files.

.travis.yml
```diff
@@ -40,7 +40,7 @@ before_script:
   - ./tests/travis_${TRANSPORT}.sh
 
 script: 
-  - phpunit -c tests/phpunit_${TRANSPORT}.xml.dist
+  - ./vendor/bin/phpunit -c tests/phpunit_${TRANSPORT}.xml.dist
```
dbu: can you fix the build in the branch of #709 please? it should also be in the maintenance branch and has nothing to do with this feature.

dantleech: This didn't fix the issue :(

tests/phpunit_doctrine_dbal.xml.dist
```diff
@@ -1,9 +1,12 @@
 <phpunit colors="true" bootstrap="bootstrap.php" backupGlobals="false" cacheTokens="true">
     <php>
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="disabled"/>
```
dantleech: May as well remove this

dbu:
can you please rebase on master and then bump the branch-alias for master to 1.4? i think apart from that we are pretty much ready.

dantleech:
Rebased and bumped the branch alias for this branch to `1.4-dev`

lib/Doctrine/ODM/PHPCR/Mapping/ClassMetadata.php
```diff
@@ -454,6 +472,57 @@ public function validateIdentifier()
...
+        if ($this->isLeaf()) {
+            throw new OutOfBoundsException(sprintf(
+                'Document "%s" has been mapped as a leaf. It cannot have children',
+                $this->name
+            ));
+        }
+
+        $childClasses = $this->getChildClasses();
+
+        if (array() === $childClasses) {
```
WouterJ: `if (0 === count($childClasses)) {` ?

tests/Doctrine/Tests/ODM/PHPCR/Mapping/ClassMetadataTest.php
```diff
@@ -324,6 +324,90 @@ public function testClassMetadataInstanceSerializationTranslationProperties($cm)
...
+     * It should return early if the given class is an allowed child class.
+     */
+    public function testAssertValidChildClassesAllowed()
+    {
+        $cm = new ClassMetadata('Doctrine\Tests\ODM\PHPCR\Mapping\Person');
+        $cm->setChildClasses(array('stdClass'));
+        $childCm = new ClassMetadata('stdClass');
+        $childCm->initializeReflection(new RuntimeReflectionService());
+        $result = $cm->assertValidChildClass($childCm);
+        $this->assertNull($result);
```
WouterJ: should be `assertTrue`, right? (same below)

WouterJ: Or the method should be changed to use `return;`. But in this case, it makes sense to test that the exception is not thrown, rather than the void return value.

dantleech: Updated to not return void instead of `true`. It should return void, so I think testing that it does return (null) does no harm or?

WouterJ: Well, there are plans for a void return type in PHP 7.x. So testing null would result in test fails if that RFC is accepted.

And probably personal, for me tests describe the public API and its behaviour. Now the tests seem to tell "it returns null if a valid document is provided". However, the real behaviour is it's throwing exceptions when something is invalid and isn't doing anything if it's valid.

dantleech: I hear that void will resolve to null. It shouldn't return anything - if it did it would be a violation of the defined public API.

dantleech:
@dbu good to merge?

tests/Doctrine/Tests/ODM/PHPCR/Mapping/ClassMetadataTest.php
dbu: a good codesniffer hopefully would complain about using the return value of a void method. i would simply delete this line. assertValidChildClass should throw an exception or no exception, and not return anything. this line could suggest to people to check the return value which makes no sense.

tests/Doctrine/Tests/ODM/PHPCR/Mapping/ClassMetadataTest.php
```diff
@@ -324,6 +324,90 @@ public function testClassMetadataInstanceSerializationTranslationProperties($cm)
...
+    /**
+     * It should return early if the given class is an allowed child class.
+     */
+    public function testAssertValidChildClassesAllowed()
+    {
+        $cm = new ClassMetadata('Doctrine\Tests\ODM\PHPCR\Mapping\Person');
+        $cm->setChildClasses(array('stdClass'));
+        $childCm = new ClassMetadata('stdClass');
+        $childCm->initializeReflection(new RuntimeReflectionService());
+        $result = $cm->assertValidChildClass($childCm);
```
dbu: i'd even remove the assignment here and just have `$cm->assertValidChildClass($childCm);`

dbu:
i'd prefer to clean up that assertNull in the tests. apart from that its ready, yes. great job!

tests/Doctrine/Tests/ODM/PHPCR/Mapping/ClassMetadataTest.php
dantleech: I don't feel very strongly about this, but we do already check for null values in this very test: https://github.com/doctrine/phpcr-odm/blob/master/tests/Doctrine/Tests/ODM/PHPCR/Mapping/ClassMetadataTest.php#L45

To me we declare that the method returns void, and we test that it returns void (i.e. null). But happy to change it.

dbu: lets change it and also change the other test then. using the result value of a void method feels wrong to me. afaik in java for example, it would be a compiler error...

dantleech:
ok, well if it could cause a compiler to crash then its not good :) removed the `assertNull` checks on the validation tests.

dbu:
yay!

dantleech:
\o/

PR #709, opened 2016-06-15
do not fail build on deprecation warnings

dbu:
build on master is failing. this is the fix from #706 ported into a separate PR, so that we can merge this first and have it in the 1.3 maintenance branch.


dbu:
@WouterJ ok so the failures is from warnings because of phpunit 5 (createMock vs getMock). but some boxes seem to use phpunit 4. yay. should we add phpunit to require-dev or download a phpunit.phar so we have the same version everywhere? moving to phpunit 5 seems like a good idea to me...

PR #710, opened 2016-06-17
Use getMockBuilder instead of getMock

dantleech:
Do not use the deprecated `getMock` method.

Fixes #709

### DIFF
--- a/tests/Doctrine/Tests/ODM/PHPCR/DocumentManagerTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/DocumentManagerTest.php
@@ -54,7 +54,7 @@ class DocumentManagerTest extends PHPCRTestCase
      */
     public function testNewInstanceFromConfiguration()
     {
-        $session = $this->getMock('PHPCR\SessionInterface');
+        $session = $this->getMockBuilder('PHPCR\SessionInterface')->getMock();
         $config = new \Doctrine\ODM\PHPCR\Configuration();
 
         $dm = DocumentManager::create($session, $config);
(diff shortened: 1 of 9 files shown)


tests/Doctrine/Tests/ODM/PHPCR/DocumentManagerTest.php
```diff
@@ -54,7 +54,7 @@ public function testFindTranslation()
      */
     public function testNewInstanceFromConfiguration()
     {
-        $session = $this->getMock('PHPCR\SessionInterface');
+        $session = $this->getMockBuilder('PHPCR\SessionInterface')->getMock();
```
dbu: the recommended way to do this is to use the new `createMock` method, afaik. the difference is that createMock disables the original constructor. only when we need additional things from the mock builder should we use getMockBuilder.

tests/Doctrine/Tests/ODM/PHPCR/UnitOfWorkTest.php
```diff
@@ -36,9 +36,9 @@ public function setUp()
         }
 
         $this->factory = new Factory;
-        $this->session = $this->getMock('Jackalope\Session', array(), array($this->factory), '', false);
+        $this->session = $this->getMockBuilder('Jackalope\Session', array()->getMock(), array($this->factory), '', false);
         
-        $this->objectManager = $this->getMock('Jackalope\ObjectManager', array(), array($this->factory), '', false);
+        $this->objectManager = $this->getMockBuilder('Jackalope\ObjectManager', array()->getMock(), array($this->factory), '', false);
```
dbu: parenthesis in the wrong place. getMockBuilder and then getMock with arguments.

dantleech: Global pearl replace error ..

tests/Doctrine/Tests/ODM/PHPCR/DocumentManagerTest.php
dantleech: The problem is that `createMock` does not exist (even in PHPUnit 5.0), and we are using, by necessity, PHPUnit 4.x in the tests with PHP 5.4

dbu:
okay, makes sense. thanks dan!

PR #711, opened 2016-06-20
start 2.0

dbu:
This is to start the 2.0 version of phpcr-odm.

Not urgent to merge into master, but we need it to release cmf 2.0 bundles because of the HierarchyInterface changes.

Nothing open to do, but tagged WIP nonetheless because we should not yet merge it.


dantleech:
dev-master of the RoutingBundle conflicts with versions < 2.0 - if we are confident that this is the right direction (I havn't got the background knowledge yet) then I would say merge it into `dev-master` alias as `2.0-dev` and remove the 2.0 branch?

dbu:
ups, you are right. i will rebase and get this ready.

dantleech:
@dbu any news on this, do you need some help?

dbu:
oh, i rebased and then forgot about it. if you have time to check that build failure, i would appreciate it. not sure if its related to the changes in here or also a problem in master.

dantleech:
Yep, broken in master

```
1) Doctrine\Tests\ODM\PHPCR\Functional\Translation\ChildTranslationStrategyTest::testQueryBuilder
Failed asserting that actual size 0 matches expected size 1.
```

dantleech:
So I would vote for merging this if you are happy with it (and removing the 2.0 branch?)

dantleech:
Oops. ignore that. The travis badge in the README is just for the last build not for master. It is this branch which is causing the failure!

dantleech:
It is failing with:

```
1) Doctrine\Tests\ODM\PHPCR\Functional\Translation\ChildTranslationStrategyTest::testQueryBuilder                      │
Failed asserting that actual size 0 matches expected size 1
```

This is because we removed the deprecated `$session->save()` from `UnitOfWork::commit()`.

So, going out on a limb, I guess the translation strategy is doing stuff directly on PHPCR which doesn't register in the changeset, so now `session->save` doesn't get called.

dbu:
unless the naming is very misleading, it sounds like an issue with test setup, not the executing. 

looking at the method, we have two calls to $strategy->saveTranslation() followed by a $dm->flush(). and the child strategy saveTranslation only updates the phpcr node without marking anything on the document. 

so either we need to revert the change and keep saving the session on flush. maybe that would be the save thing to do. or we need a way for the translation strategy to mark the document as dirty so the uow triggers the session save even though it does not detect changes on the document itself.

how much of a problem is it if $dm->flush keeps saving an "unchanged" phpcr session, thus flushing phpcr changes not seen by the dm? while its kind of unclean to do this, i feel the wtf factor is quite high when saving my phpcr changes depends on whether randomly a document also changed or not - as in phpcr we can't save partial changes, only the whole session.

@lsmith77 do you have an opinion?

wouterj:
Hi guys, just thought about this package: All bundles depend on this one, so we have to release 2.0.0-stable (and the RC) as well this month. What's left to do here?

dbu:
looking at https://github.com/doctrine/phpcr-odm/compare/1.4.0...master i wonder if we should not make the cmf bundles require ^1.4.0 and when we have more things on phpcr-odm and release a 2.0, allow 1.4|2.0 ?

PR #713, opened 2016-07-20
Register system node types in dedicated class

### LINKED ISSUE #712: Decouple registration of system types from the command.
Currently the system types are defined within the console command, meaning that in order to add these types pragmatically it is necessary either to instantiate the command or copy the node definitions.

dantleech:
Fixes #712 

This is a quick fix to decouple the registration of system types from the console command.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/NodeTypeRegistrator.php
+++ b/lib/Doctrine/ODM/PHPCR/NodeTypeRegistrator.php
@@ -1,0 +1,57 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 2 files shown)


lib/Doctrine/ODM/PHPCR/NodeTypeRegistrator.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Encapsulates the logic for registering system node types.
+ */
+final class NodeTypeRegistrator
+{
+    private $phpcrNamespace = 'phpcr';
+    private $phpcrNamespaceUri = 'http://www.doctrine-project.org/projects/phpcr_odm';
+    private $localeNamespace = Translation::LOCALE_NAMESPACE;
+    private $localeNamespaceUri = Translation::LOCALE_NAMESPACE_URI;
```
dantleech: We could've made these constructor arguments, but currently the design does not permit changing these values, so would be more work.

dbu: should we make them constants to make it more obvious they are not changed?

dbu:
thanks dan! can you please rebase this branch? i let styleci fix the master branch.

lib/Doctrine/ODM/PHPCR/NodeTypeRegistrator.php
dantleech: Issue would be that then people could then reference them - but they should imo be configurable values.

dantleech:
rebased!

lib/Doctrine/ODM/PHPCR/NodeTypeRegistrator.php
dbu: okay, makes sense.

PR #716, opened 2016-07-23
Cascade children

dantleech:
Support for cacading children.

More work probably needs to be done, but quite tired so will just leave this here :)

This allows:

```php

$child1 = new Child();
$child2 = new Child();

$document->myChildren = [ $child1, $child2 ];
$dm->persist($document);
$dm->flush();
```


dantleech:
Hmm, for some reason my use case works also without this fix. Closing.

PR #717, opened 2016-08-03
Temporary revert getField() removal

wouterj:
Some slow moving vendor dependencies (like JMS Serializer, ref
schmittjoh/serializer#605) hold up the development at the moment. Keeping
this method for the moment will fix it.

This PR should be reverted before 2.0.0 is released.


wouterj:
build failures are not related

dbu:
thanks. we might add a comment to the deprecated annotation in the code linking to the jms serializer pr, so we can quickly check if its still needed in the future.

PR #721, opened 2016-09-22
Use master branch for travis branch

dantleech:
Use the master branch for the travis badge.


dantleech:
OK, right for the wrong reason. The old badge was pointing to `master`, but it seems travis changed the URL format. So this is correct (and shows the `master` branch as passing)

dbu:
thanks. it looks like the phpbench run is now failing on master. but thats not related to this change.

dantleech:
Hmph. Yeah regression with v12.0 :/

https://github.com/phpbench/phpbench/issues/446

Locking to 11.0 would fix the issue, or I will probably make a point
release to fix the issue this week.

dantleech:
Fixed PHPBench

Issue #723, opened 2016-10-01
Xml Driver "child-class" mapping returns SimpleXml instead of string.


dantleech:
```
$childClasses = $metadata->getChildClasses();
```

```
array (size=2)
  0 => 
    object(SimpleXMLElement)[576]
      public 0 => string 'Foo\Bundle\Media\Document\File' (length=31)
  1 => 
    object(SimpleXMLElement)[570]
      public 0 => string 'Foo\Bundle\Media\Document\Folder' (length=33)
```

dbu:
fixed in #724

PR #724, opened 2016-10-01
Cast child-class XML attributes to strings

dantleech:


dantleech:
Fixes #723

dbu:
thanks. kind of worrying that our tests did not catch that. we do have tests for xml mapping in general, just missed the child mapping?

dantleech:
I am pretty sure we test the XML driver - but I think the XML attribute
element gets cast to a string during assertion.

PR #730, opened 2016-12-16
Do not refresh session on UnitOfWork::clear()

pamil:
See discussion in doctrine/DoctrinePHPCRBundle#265.

###### Contains and resolves #731. Fixes doctrine/DoctrinePHPCRBundle#265.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -3201,8 +3201,6 @@ class UnitOfWork
         $this->documentVersion = array();
 
         $this->invokeGlobalEvent(Event::onClear, new OnClearEventArgs($this->dm));
-
-        $this->session->refresh(false);
     }
 
     /**


dantleech:
As mentioned in the [issue](https://github.com/doctrine/DoctrinePHPCRBundle/issues/265), I am not 100% sure this is the correct thing to do.

pamil:
The ORM does not refresh entities on clear as well, so we may use the same pattern here.

Anyway, refreshing documents on clearing the `UnitOfWork` does not sound useful.

pamil:
The Travis failure is not connected with this PR 😿

dantleech:
@dbu @lsmith77 any idea why the session is refreshed when `UnitOfWork::clear` is called?

pamil:
It was introduced in https://github.com/doctrine/phpcr-odm/commit/ce8fd9274875ef1e019bfaac1c5f9e931f7b1e7b, #67.

pamil:
Rebased.

dbu:
the difference is that the orm will do new sql queries when the object store is cleared. phpcr has an object model of nodes and properties that stays in memory even if we delete the odm objects. depending on the use case, one or the other is the correct thing to happen.

the stack trace in https://github.com/doctrine/DoctrinePHPCRBundle/issues/265 is a bug in jackalope - during refresh, it should only mark nodes as dirty so that they get reloaded from storage when accessed again. i created https://github.com/jackalope/jackalope/issues/321 about that.

while i think we could make a case for not discarding changes on phpcr level, i fear there is a BC break in this. people that expect it to happen would be confused. i think the best solution at this point is adding a flag to the DocumentManager::clear method whether to discard the phpcr session as well or not. something like `$refreshPhpcr = true`. in the place you call clear, you can specify false to get the behaviour you need.

dantleech:
> phpcr has an object model of nodes and properties that stays in memory even if we delete the odm objects.

but at the ODM level - any changes in the UOW are not made on the PHPCR level until `flush()` is called. So clearing the UOW should not affect PHPCR at all? (?)

dbu:
question is whether you want to clear to just discard your changes or also to see changes that happened at the backend since you did your thing. e.g. a recovery code after an exception during flush that something got deleted at the backend - if you clear and try again, phpcr-odm will still see the outdated phpcr nodes and things wont be fixed...

we can discuss changing the default behaviour for clear(), but i would keep both modes. and to fix the issue at hand, having at least an option to tell not to refresh seems good.

dbu:
so, 2.0 is near, so that would be a good time for behaviour changes.

one option i see is stop refreshing the phpcr layer and document that if you want to reset the phpcr level cache, you need to refresh the session. i would kind of like to rename the method so that people don't leave `clear` calls in their code without realizing that the semantics changed. but `clear` is the name also used in orm, and seems the best name for what we do.

@lsmith77 do you maybe have some input here?

dantleech:
> i would kind of like to rename the method so that people don't leave clear calls in their code without realizing that the semantics change

I don't think `clear` suggests `reload from database`, so would rather keep that name than make a compromise. The break should be clearly documented and I expect most people will make a concious decision to updade to 2.0 unless they are depending on `dev-master`.

pamil:
I'd vote for making this `clear()` method consistent with the ORM (it's the most expected behaviour IMO) and allow for refreshing session some other way (`$unitOfWork->refresh();` - making the parameter optional?).

PR #733, opened 2017-01-15
implement getDocumentId

### LINKED ISSUE #597: getDocumentId in DocumentManager?
It is often required that you need to get a document path without knowing about any accessors the document may have (if any) for that purpose.

Typically this happens when you are implementing event handlers.

Currently it is necessary to call `$documentManager->getUnitOfWork()->getDocumentId()`. The document manager already has calls which I think are on the same level, such as `DocumentManager#getDocumentChildren()`, `DocumentManager#getReferrers`, etc.

So I propose making this method part of the document manager somehow, maybe there should be two methods, e.g. `getDocumentUuid`, `getDocumnetPath`.

ElectricMaxxx:
fix #597

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/DocumentManager.php
+++ b/lib/Doctrine/ODM/PHPCR/DocumentManager.php
@@ -925,4 +925,12 @@ class DocumentManager implements DocumentManagerInterface
 
         return $this->session->getNode($path);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDocumentId($document)
+    {
+        return $this->unitOfWork->getDocumentId($document);
+    }
 }
--- a/lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php
+++ b/lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php
@@ -525,4 +525,15 @@ interface DocumentManagerInterface extends ObjectManager
      * @throws PHPCRException                if $document is not managed
      */
     public function getNodeForDocument($document);
+
+    /**
+     * Get the document identifier phpcr-odm is using.
+     *
+     * @param object $document A managed document
+     *
+     * @return string
+     *
+     * @throws PHPCRException if $document is not managed
+     */
+    public function getDocumentId($document);
 }
--- a/tests/Doctrine/Tests/ODM/PHPCR/Decorator/DocumentManagerDecoratorTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Decorator/DocumentManagerDecoratorTest.php
@@ -2,6 +2,8 @@
 
 namespace Doctrine\Tests\ODM\PHPCR\Decorator;
 
+use Doctrine\ODM\PHPCR\Decorator\DocumentManagerDecorator;
+use Doctrine\ODM\PHPCR\DocumentManager;
 use Doctrine\Tests\ODM\PHPCR\PHPCRTestCase;
 
 /**
@@ -9,6 +11,8 @@ use Doctrine\Tests\ODM\PHPCR\PHPCRTestCase;
  */
 class DocumentManagerDecoratorTest extends PHPCRTestCase
 {
+    /**
+     */
     public function testCheckIfAllPublicMethodsAreDecorated()
     {
         $dmMethods = get_class_methods('Doctrine\ODM\PHPCR\DocumentManager');
@@ -19,11 +23,15 @@ class DocumentManagerDecoratorTest extends PHPCRTestCase
         $dmiMethods = array_diff($dmiMethods, array('__construct'));
         sort($dmiMethods);
 
-        $dmdMethods = get_class_methods('Doctrine\ODM\PHPCR\\Decorator\DocumentManagerDecorator');
+        $dmdMethods = get_class_methods('\Doctrine\Tests\ODM\PHPCR\Decorator\OwnDocumentManager');
         $dmdMethods = array_diff($dmdMethods, array('__construct'));
         sort($dmdMethods);
 
         $this->assertEquals($dmMethods, $dmiMethods);
         $this->assertEquals($dmMethods, $dmdMethods);
     }
+}
+
+class OwnDocumentManager extends DocumentManagerDecorator
+{
 }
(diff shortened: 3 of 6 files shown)


ElectricMaxxx:
@dbu ping, is there no travis?

composer.json
```diff
@@ -21,7 +21,8 @@
         "phpcr/phpcr-implementation": "^2.1.0",
         "phpcr/phpcr-utils": "^1.2.8",
         "doctrine/instantiator": "^1.0.1",
-        "symfony/console": "^2.3|^3.0"
+        "symfony/console": "^2.3|^3.0",
+        "jackalope/jackalope-doctrine-dbal": "~1.0"
```
ElectricMaxxx: arg, sorry will remove that.

lib/Doctrine/ODM/PHPCR/DocumentManager.php
```diff
@@ -925,4 +925,16 @@ public function getNodeForDocument($document)
...
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDocumentId($document)
+    {
+        try {
+            return $this->unitOfWork->getDocumentId($document);
+        } catch (PHPCRException $e) {
```
dbu: why eat away the exception? i think i would prefer to have exceptions here as well, and let the user handle them. otherwise people will be confused why the id is null when they ask for the id of the wrong object.

ElectricMaxxx: I think different. When doing `find()` we also eat several exceptions. When having a new document which isn't managed yet,  a null as ID sounds valid. Just the "not managed due to problems" aspects would confuse with null as a return value.

dbu: getLocalesFor would be an example where we get an exception on an unmanaged document: https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/DocumentManager.php#L546 

`find` indeed does return null on not found. but i feel that is a design flaw of doctrine - i'd rather have an exception in that case. but i don't want to change that now, it would be confusing when orm does return null instead of ecxeption. however, new methods can be done in a more secure way.

ElectricMaxxx:
:+1: OK, will change sounds valid.

ElectricMaxxx:
@dbu it is also implemented in `\Doctrine\ODM\PHPCR\Decorator\DocumentManagerDecorator`, which is an abstract class, which would force extending classes to implement that method now. As i don't know what the usecase of that decorator is, i don't know should let the abstract class implement that method or force the extending class to do so.

dbu:
the idea of the decorator is that if you just one a wrapper around DocumentManager that implements DocumentManagerInterface but usually forwards calls to DocumentManager, you don't need to implement all the methods that you don't want to change. so yes, please add the method on the decorator too.

ElectricMaxxx:
i will do so, but i am a little bit confused, why the test `\Doctrine\Tests\ODM\PHPCR\Decorator\DocumentManagerDecoratorTest` did not fail.

dbu:
indeed, that is surprising and a bit worrying. if you want to debug, check if the test is even executed, then dump the method lists to see if they look sane. it must be something stupid i guess.

ElectricMaxxx:
I think `abstract class Bla implement Blub` lies when asking for `get_class_methods`

ElectricMaxxx:
Argh and even Reflection seems to lie. :-(

lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php
```diff
@@ -525,4 +525,15 @@ public function close();
      * @throws PHPCRException                if $document is not managed
      */
     public function getNodeForDocument($document);
+
+    /**
+     * Evaluates the documents identifier. Will return null when document is not managed.
```
dbu: the "will return null when document is not managed" is no longer correct. i would remove it and rely on the `@throws` annotation.

maybe say: "Get the document identifier phpcr-odm is using."

lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php
```diff
@@ -525,4 +525,15 @@ public function close();
...
+     *
+     * @param object $document
```
dbu: lets add the description: "A managed document"

tests/Doctrine/Tests/ODM/PHPCR/DocumentManagerTest.php
```diff
@@ -165,6 +166,35 @@ public function testCreateQueryBuilder()
...
+
+        $obj = new \stdClass;
+        $uow = $dm->getUnitOfWork();
+
+        $reflectionProperty = new \ReflectionProperty($uow, 'documentIds');
+        $reflectionProperty->setAccessible(true);
+        $reflectionProperty->setValue($uow, array(spl_object_hash($obj) => '/foo'));
+        $reflectionProperty->setAccessible(false);
+
+
```
dbu: can you eliminate the extra blank line please?

dbu:
interesting. but i agree on the fix if that one works better. if we add anything to the interface, we immediately see the problem.

ElectricMaxxx:
I fixed and squashed, ready to merge from my POV.

tests/Doctrine/Tests/ODM/PHPCR/Decorator/DocumentManagerDecoratorTest.php
```diff
@@ -2,13 +2,17 @@
...
+use Doctrine\ODM\PHPCR\DocumentManager;
 use Doctrine\Tests\ODM\PHPCR\PHPCRTestCase;
 
 /**
  * @group unit
  */
 class DocumentManagerDecoratorTest extends PHPCRTestCase
 {
+    /**
+     */
```
dbu: tests need no comments, and this one is empty. please remove this

dbu:
please remove the empty comment block on the test case, then i am fine if you just merge it. thanks!

ElectricMaxxx:
@dbu did it, but i have no rights on this repository ;-)

dbu:
thanks a lot max!

PR #734, opened 2017-01-17
references by uuid

ElectricMaxxx:
This should wrap up #493 

didn't implement the stuff, commented in code not to change.

But there are still the [Questions](https://github.com/doctrine/phpcr-odm/pull/493#issuecomment-43854355) of @dbu

I go top to button through the text/conversation, cause it is really hard to remember what the purpose was, even for me :-)

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/DocumentManager.php
+++ b/lib/Doctrine/ODM/PHPCR/DocumentManager.php
@@ -917,13 +917,11 @@ class DocumentManager implements DocumentManagerInterface
      */
     public function getNodeForDocument($document)
     {
-        if (!is_object($document)) {
-            throw new InvalidArgumentException('Parameter $document needs to be an object, '.gettype($document).' given');
+        if (!$identifier = $this->unitOfWork->getDocumentId($document)) {
+            throw new InvalidArgumentException('This document is not managed by this manager.');
         }
 
-        $path = $this->unitOfWork->getDocumentId($document);
-
-        return $this->session->getNode($path);
+        return $this->unitOfWork->getNodeByPathOrUuid($identifier);
     }
 
     /**
--- a/lib/Doctrine/ODM/PHPCR/DocumentManager.php
+++ b/lib/Doctrine/ODM/PHPCR/DocumentManager.php
@@ -280,29 +280,28 @@ class DocumentManager implements DocumentManagerInterface
      */
     public function find($className, $id)
     {
+        if (!UUIDHelper::isUUID($id) && strpos($id, '/') !== 0) {
+            $id = '/'.$id;
+        }
+
+        // try to get a manage document
+        $document = $this->unitOfWork->getDocumentByPathOrUuid($id);
+        if ($document) {
+            try {
+                $this->unitOfWork->validateClassName($document, $className);
+
+                return $document;
+            } catch (ClassMismatchException $e) {
+                return null;
+            }
+        }
+
+        // get the node
         try {
-            if (UUIDHelper::isUUID($id)) {
-                try {
-                    $id = $this->session->getNodeByIdentifier($id)->getPath();
-                } catch (ItemNotFoundException $e) {
-                    return null;
-                }
-            } elseif (strpos($id, '/') !== 0) {
-                $id = '/'.$id;
-            }
-
-            $document = $this->unitOfWork->getDocumentById($id);
-            if ($document) {
-                try {
-                    $this->unitOfWork->validateClassName($document, $className);
-
-                    return $document;
-                } catch (ClassMismatchException $e) {
-                    return null;
-                }
-            }
-            $node = $this->session->getNode($id);
+            $node = $this->unitOfWork->getNodeByPathOrUuid($id);
         } catch (PathNotFoundException $e) {
+            return null;
+        } catch (ItemNotFoundException $e) {
             return null;
         }
 
@@ -368,7 +367,7 @@ class DocumentManager implements DocumentManagerInterface
                 $id = '/'.$id;
             }
 
-            $document = $this->unitOfWork->getDocumentById($id);
+            $document = $this->unitOfWork->getDocumentByPathOrUuid($id);
 
             if ($document) {
                 $this->unitOfWork->validateClassName($document, $className);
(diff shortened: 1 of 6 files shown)


lib/Doctrine/ODM/PHPCR/DocumentManager.php
```diff
@@ -917,13 +917,11 @@ public function initializeObject($document)
      */
     public function getNodeForDocument($document)
     {
-        if (!is_object($document)) {
```
dantleech: Why do we remove this validation?

lib/Doctrine/ODM/PHPCR/DocumentManager.php
```diff
@@ -917,13 +917,11 @@ public function initializeObject($document)
...
-            throw new InvalidArgumentException('Parameter $document needs to be an object, '.gettype($document).' given');
+        if (!$identifier = $this->unitOfWork->getDocumentId($document)) {
```
dantleech: `if (null === $identifier = $this->...`

lib/Doctrine/ODM/PHPCR/DocumentManager.php
```diff
@@ -917,13 +917,11 @@ public function initializeObject($document)
...
+            throw new InvalidArgumentException('This document is not managed by this manager.');
```
dantleech: `sprintf('Document of class "%s" is not managed', get_class($document))`

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -674,7 +675,7 @@ public function getOrCreateProxy($targetId, $className, $locale = null)
         $metadata      = $this->dm->getClassMetadata($className);
         $proxyDocument = $this->dm->getProxyFactory()->getProxy($className, array($metadata->identifier => $targetId));
 
-        // register the document under its own id
+        // register the document under its own id/uuid
```
dantleech: `path/uuid`.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3113,6 +3126,20 @@ public function getDocumentById($id)
     }
 
     /**
+     * Creates a node from a given path or an uuid
```
dantleech: `given path or UUID`

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3063,18 +3069,25 @@ private function unregisterDocument($document)
...
 
         // frozen nodes need another state so they are managed but not included for updates
-        $frozen = $this->session->nodeExists($id) && $this->session->getNode($id)->isNodeType('nt:frozenNode');
+        try {
+            $node = $this->getNodeByPathOrUuid($pathOrUuid);
+            $frozen = $node instanceof  NodeInterface ? $node->isNodeType('nt:frozenNode') : false;
+        } catch (ItemNotFoundException $e) {
+            $frozen = false;
+        } catch (PathNotFoundException $e) {
+            $frozen = false;
```
dantleech: `$frozen = false` can be moved before the `try` block (doesn't need to be repeated twice)

tests/Doctrine/Tests/ODM/PHPCR/UnitOfWorkTest.php
```diff
@@ -74,54 +75,66 @@ public function setUp()
...
             ->with($id)
             ->will($this->returnValue(true));
-        
+
         $nodeData = array(
             "jcr:primaryType" => $primaryType,
             "jcr:system" => array(),
             'username' => $username,
         );
+        if (null !== $uuid) {
```
dantleech: I prefer a blank line before most if statements.

tests/Doctrine/Tests/ODM/PHPCR/UnitOfWorkTest.php
```diff
@@ -74,54 +75,66 @@ public function setUp()
...
-        
+
+        if (null !== $uuid) {
+            $this->session->expects($this->any())
+                ->method('getNodeByIdentifier')
+                ->with($uuid)
+                ->will($this->returnValue($node));
+        }
+
+
```
dantleech: One blank line too many

dantleech (commented):
Can you summarize the motivation behind this PR in the description? I don't really know the full implications of this change.

ElectricMaxxx:
This will fail for sure now ... Did a lot of changes

* introduced a object hash - uuid map, which is filled when a document is registered by its uuid (on `DocumentManager::getReference()` call with proxies referencing documents by their uuid <- @dantleech purpose)
* introduced a uuid - path map to find documents by their path in identity map, which are managed by uuid

so all documents still live in `identityMap` by their id, but the id can be the absolute paths directly or uuids which needs to be converted into paths (uuid - id map).

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -100,6 +101,18 @@ class UnitOfWork
     private $documentIds = array();
 
     /**
+     * @var array
+     */
+    private $documentUuids = array();
```
dantleech: Ideally this would be better named `$oidToUuidMap`, but to be consistent this naming is fine for now I guess. But it should be commented.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -100,6 +101,18 @@ class UnitOfWork
...
+
+    /**
+     * Mapping to find ids (abs. path) of document managed by its uuid.
+     *
+     * @var array
+     */
+    private $uuidIdMapping = array();
```
dantleech: `Id` in `uuidId` is redundant. Better to name `$uuidToPathMap`.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -3058,23 +3076,43 @@ private function unregisterDocument($document)
             $this->documentVersion[$oid]
         );
 
+        if (!empty($this->documentUuids[$oid]) && array_key_exists($this->documentUuids[$oid], $this->uuidIdMapping)) {
```
dantleech: Why check for `!empty`?

dantleech (commented):
Still not sure what the *motivation* of this is? Regarding the mapping I wonder if it is worth introducing a Registry class to encapsulate the various maps, reducing complication in this already very complicated class, something *like*: https://github.com/sulu/sulu-document-manager/blob/master/lib/DocumentRegistry.php.

But maybe something for a later iteration.

PR #735, opened 2017-01-31
Don't override ID generator when persisting children

wouterj:
When persisting a new document with children, the children are automatically persisted as well. However, this persistence of the children is done using an override ID generator (assigned). This means the children are required to have an `@Id` field in order to be persisted.

I don't know why this is done and it didn't work as expected when working with the PHPCR ODM (I don't want to use the assigned ID strategy).

The line is added in https://github.com/doctrine/phpcr-odm/pull/50 . Unfortunately, the related issue can no longer be read as Jira is removed. However, the tests all still pass, so I don't think I've broken the fix.

Again, I'm not an PHPCR ODM expert and don't know if there was a specific reason to do this, but using the one from the ClassMetadata makes a lot more sense in my head.

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -1689,7 +1689,7 @@ class UnitOfWork
                 if ($this->getDocumentById($childId)) {
                     $child = $this->merge($child);
                 } else {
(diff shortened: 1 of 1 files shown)


dbu:
interesting that this is named the same as #50 :-/

but i think you are right. looking at the id strategies, if the id is assigned already, they all behave like assigned id anyways. so if you have the id field, nothing should change. if you don't have the id field, you might get lucky to get something persisted. the ParentId strategy for example will only work if the document name has been set already.

dbu:
but this is definitely an improvement. thanks for the fix!

Issue #737, opened 2017-02-16
NodeName is not the same like Nodename (especially in case sensitive file systems)


stopsopa:
File https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/Mapping/Annotations/Nodename.php

is named "Nodename.php"
but according documentation in annotation we should use:
@PHPCR\NodeName

http://docs.doctrine-project.org/projects/doctrine-phpcr-odm/en/latest/reference/annotations-mapping.html#note-on-usage

but in linux environment which have case sensitive file system it cause error:
http://i.imgur.com/anpMybN.png

ODM is great :)

Cheers

tomekwlod:
Same here. MacOS is OK, Vagrant under MacOS is still fine, but Centos7 is already crashing

dbu:
i think the mistake is in the documentation. we use Nodename elsewhere: http://docs.doctrine-project.org/projects/doctrine-phpcr-odm/en/latest/reference/annotations-mapping.html#nodename

looking at the codebase, we are super inconsistent in naming with this. many method names use NodeName. even the annotation is mixed up in some tests. which is surprising me, because travis-ci is running on linux architectures and i work on that as well.

given that if the name is case sensitive, its a BC break to change this to uppercase, I vote that we instead fix all documentation to be consistent on Nodename and fix the tests to always have lowercase name.

stopsopa:
IMO shoud be NodeName because next to it you're using @PHPCR\ParentDocument().
According to You thinking this should also be
@PHPCR\Parentdocument()

stopsopa:
.. yes it's strange that travis-ci didn't caught that

dbu:
i agree that Nodename is not nice. probably the right thing would be to add NodeName alongside and fix the code to handle both camelcase and the wrong one. that would be BC, and we can deprecate the Nodename thing and remove it in 2.0. but i have no idea what a case insensitive filesystem will do if our repository provides both files... would you happen to know?

PR #739, opened 2017-02-16
NodeName is not the same like Nodename

stopsopa:
Related to: https://github.com/doctrine/phpcr-odm/issues/737


dbu:
sorry, i prefer to go with consistent Nodename over risking a BC break for people on case sensitive systems. see #761

PR #749, opened 2017-08-01
Fix updating new Id for children after a move.

UFTimmy:
The assignment was using the object id of the parent, instead of the child.


dbu:
indeed, thanks a lot!

if you have the time, it would be great if you can add checks to the [move test](https://github.com/doctrine/phpcr-odm/blob/defb06ee92a4252ad0f24584fb3ac820f112ba01/tests/Doctrine/Tests/ODM/PHPCR/Functional/MoveTest.php#L135) to ensure that this keeps working.

PR #758, opened 2017-09-11
drop php 5 support for version 2

dbu:
reduced to only drop 5.x and 7.0 support and add 7.1 testing. adding php 7.2 support is blocked by the dependencies needing to support it first. will be done in a separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,7 @@
 language: php
 
 php:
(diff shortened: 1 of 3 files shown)


dbu:
7.0 works in master, but fails here. i restarted the build on master, and its still fine. can't spot the difference of what i am doing!

PR #759, opened 2017-09-11
test with php 7.2

dbu:
blocked by https://github.com/phpcr/phpcr-utils/pull/174 until 7.2 RC6 is out


dbu:
closing in favor of #765 - we should fix this in the 1.4 branch as it should be trivial changes only

PR #760, opened 2017-11-06
Use PSR-1 for PHPUnit TestCase

gabriel-caruso:
Use [PSR-1](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names) while extending PHPUnit TestCase class. This will help us when to migrate to PHPUnit 6, that [no longer support snake case namespaces](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "symfony/yaml": "^3.3",
         "symfony/phpunit-bridge": "^3.2",
         "liip/rmt": "~1.2",
-        "phpunit/phpunit": "5.*"
+        "phpunit/phpunit": "~5.7.21"
     },
     "suggest": {
         "symfony/yaml": "^3.2",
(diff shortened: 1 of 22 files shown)


dbu:
thanks a lot for tackling this issue!

in #759 i try to drop php 5 for version 2.0 of phpcr-odm. with that it will become easier to just jump to phpunit 6.

note that master is still about phpcr-odm 1.4 - i think i will create a 1.4 branch and move 2.0 to become master. but 2.0 has one test that fails, confusingly.

gabriel-caruso:
@dbu Only `PHP 5.3` test isn't passing. This version isn't supported for PHP anymore, neither PHPUnit. Should I drop from our tests?

dbu:
i am trying to drop old versions in #711 - there is one odd error happening with php 7.0 and 7.1. it smells like its related to some of the other cleanups in the 2.0 branch because 1.4 builds successfully on 7.0

gabriel-caruso:
@dbu I'll work on that. I give updates ASAP

dbu:
only a small thing left for #711 - lowest version build. i'll figure that out i hope.

gabriel-caruso:
I'll rebase this PR, sorry for the delay.

gabriel-caruso:
I've also bumped PHPUnit version to [5.7.21](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#5721---2017-06-21), that contains support for these new `namespaces`.

dbu:
great, thanks!

PR #765, opened 2017-11-22
Test with PHP 7.1 and 7.2

dbu:


ElectricMaxxx:
Is the `each` somewhere at our side?

dbu:
i have no idea, would be really helpful if there was a stack trace or at least the file where this happened. not sure why it does not show. tried finding it locally but don't find it anymore.

ElectricMaxxx:
@dbu you could run the the test with a regex on deprecation helper:
https://symfony.com/blog/new-in-symfony-3-1-deprecation-helper-improvements
the result should serve a stacktrace then.

PR #768, opened 2017-12-06
Update master from 1.4

dbu:


dbu:
gna. so the latest phpunit 5 still has the each problem, or its coming from somewhere else.

i guess i ignore that for now and try to upgrade to phpunit 6 separately

.travis.yml
```diff
@@ -10,39 +11,47 @@ cache:
     - $HOME/.composer/cache/files
 
 env:
-  - TRANSPORT=jackrabbit PACKAGE_VERSION=high
-  - TRANSPORT=doctrine_dbal PACKAGE_VERSION=high
+  global:
+    # on PHP 7.2 we get deprecations about using each. they supposedly come from phpunit 5 itself. TODO: update to phpunit 6 and see if this goes away.
```
alexislefebvre: Maybe you already know this but `each` is deprecated in PHP 7.2:

> **Warning** This function has been DEPRECATED as of PHP 7.2.0. Relying on this function is highly discouraged.

Source: http://php.net/manual/en/function.each.php

So I don't think that the problem comes from PHPUnit. :wink:

dbu: thanks for pointing that out, but i know that PHP 7.2 deprecated `each`. i hunted down the each-es in this library and the dependencies, but then was left with warnings for each without indication of where it happens. and somebody suggested it could be phpunit itself. now that i am upgrading to phpunit 6, the warning goes away indeed, so i think phpunit 4 and 5 use `each` internally somewhere.

PR #769, opened 2017-12-06
do not ignore 'each' deprecation and upgrade to phpunit 6

dbu:
Adjust to phpunit 6
* [x] Rewrite getMock to createMock where possible
* [x] expectException instead of setExpectedException

Other cleanup
* [x] Use ::class pseudo-constants

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,6 @@ cache:
     - $HOME/.composer/cache/files
 
 env:
-  global:
-    # on PHP 7.2 we get deprecations about using each. they supposedly come from phpunit 5 itself. TODO: update to phpunit 6 and see if this goes away.
-    - SYMFONY_DEPRECATIONS_HELPER="/.*each.*/"
   matrix:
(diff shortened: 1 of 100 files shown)


dbu:
@carusogabriel just before you start working on phpcr-odm: there is this WIP to cleanup the code and move to phpunit 6. (and yes, its doing too many things in one PR)

carusogabriel:
Ok! Do you want some help to do it? I don’t know how WIPs work, but I can help

dbu:
for now its just a codemonkey thing (and figuring out what i broke when doing large search-replaces). i will ping you if i need help, thanks. but for now i just saw your other pull requests on jackalope and phpcr and wanted to make sure you don't repeat the work i do here ;-)

dbu:
yay, here we go. i am sure there is a bunch of smaller tweaks that could be done in the tests. like avoiding the warnings that the mapping tests don't have any assertions. (essentially they try to load something and if it does not fail it means it worked...)

PR #771, opened 2017-12-13
improve styleci config

dbu:
we are now on php 7 and removed incompatible classes.

do we want to stay on the psr2 preset, or should we use the symfony preset or something else? (doctrine orm does not use styleci at all but phpcs directly)

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,12 +1,1 @@
-preset: psr2
-
-enabled:
-  - long_array_syntax
-  - duplicate_semicolon
-
-finder:
-  not-name:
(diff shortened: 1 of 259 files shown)


dbu:
hm, not happy with the psr2 defaults. i would have hoped to at least get short array notation...

we could go with `symfony` or `recommended`: https://styleci.readme.io/v1.0/docs/presets

dbu:
@ElectricMaxxx https://styleci.io/analyses/Xay5Wm would show the changes with the recommended style. i like that, you too?

ElectricMaxxx:
Looks good to me. I really like the no space philosophy for concatenation. Alphabeticial order for use statements is not that imporrtant for me but it is ok.

PR #794, opened 2019-10-10
Add link to doctrine 2 architecture in the documentation

Pictor13:
Points to latest version of the docs. Better that than a specific version that'll be outdated in the future.
The user will be shown the version number and will be warned anyway if the version is not yet released or too old.


dbu:
thanks!

PR #795, opened 2019-10-15
fix translations missing when using the query builder with joins on referrer fields

esserj:
I'm not entirely sure why it's the document manager findMany is failing here, but the following scenario fails:

I have a `project.category` and a `category.projects`, where `category.projects` is a referrer to a project document referenced by its `category` field and category and project both have translatable fields.
when u then write the following query:

```php
$queryBuilder->fromDocument(ProjectCategory::class, 'pc');
$join = $queryBuilder->addJoinInner();
$join->right()->document(Project::class, 'p');
$join->condition()->equi('p.category', 'pc.projects');
```

It will return all categories, but without translations mapped to it

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/DocumentManager.php
+++ b/lib/Doctrine/ODM/PHPCR/DocumentManager.php
@@ -320,6 +320,10 @@ class DocumentManager implements DocumentManagerInterface
      */
     public function findMany($className, array $ids)
     {
+        // when loading duplicate ID's the resulting response would be a collection of unique ids,
+        // but having duplicates would also cause a lot of overhead as well as break translation loading,
+        // so pre filter unique, see https://github.com/doctrine/phpcr-odm/pull/795
+        $ids = array_unique($ids);
         $uuids = array();
         foreach ($ids as $key => $id) {
             if (UUIDHelper::isUUID($id)) {
--- a/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/DocumentManagerTest.php
+++ b/tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/DocumentManagerTest.php
@@ -2,15 +2,17 @@
 
 namespace Doctrine\Tests\ODM\PHPCR\Functional\Translation;
 
-use Doctrine\Tests\Models\Translation\Article;
-use Doctrine\Tests\Models\Translation\Comment;
-use Doctrine\Tests\Models\Translation\InvalidMapping;
-use Doctrine\Tests\Models\Translation\DerivedArticle;
-use Doctrine\Tests\Models\CMS\CmsArticle;
-use Doctrine\Tests\ODM\PHPCR\PHPCRFunctionalTestCase;
 use Doctrine\ODM\PHPCR\Mapping\ClassMetadata;
-use Doctrine\ODM\PHPCR\Translation\LocaleChooser\LocaleChooser;
 use Doctrine\ODM\PHPCR\PHPCRException;
+use Doctrine\ODM\PHPCR\Translation\LocaleChooser\LocaleChooser;
+use Doctrine\Tests\Models\CMS\CmsArticle;
+use Doctrine\Tests\Models\Translation\Article;
+use Doctrine\Tests\Models\Translation\ArticleCategory;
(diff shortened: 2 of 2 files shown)


dbu:
thank you for reporting this and looking into it!

i realize we never released version 2. probably we should just do that with whatever we have, rather than wait for more things that don't actually happen. the build of version 1 has fallen victim to code rot, i guess most of the travis-ci setup for those old build images is not good anymore...

could you change this to be against the master branch?

it would be great to have a test to reproduce the problem you describe.

i am not sure if the solution you propose adresses the problem correctly. if fallback is broken when using findMany, that would affect other places besides querying. if you do findMany with a list of ids, is the translation system then also not working?

esserj:
Hi @dbu the solution I propose here is fiing the problem, but as you already suggested there is another underlying cause I found during implementing tests:

The problem I have is only caused when there are multiple projects with the same category.
In this case the $ids array has duplicates for each category based on how many projects it has.

so 3 categories with each 3 projects will hold 9 $ids.
This then somehow causes the problem, and thats why my solution fixed it because it filtered out the duplicates and only asked it one, I'm busy further investigating the problems root cause and will target both branches with a fix and tests to prove it

lib/Doctrine/ODM/PHPCR/DocumentManager.php
```diff
@@ -477,13 +477,23 @@ public function getDocumentsByPhpcrQuery(QueryInterface $query, $className = nul
...
+                $hints = array('fallback' => true);
+                $documents[$id] = $this->getUnitOfWork()->getOrCreateDocument(
+                    $className,
+                    $row->getNode($primarySelector),
+                    $hints
+                );
+            }
         }
 
-        return $this->findMany($className, $ids);
```
dbu: if the problem is duplicate ids, could we `array_unique($ids)` somewhere? findMany does getOrCreateDocuments with the list of nodes - as far as i can see, getOrCreateDocuments does not check if we have duplicates, but does some refresh if asked to, and other funny stuff (ouch, that method is complicated). imho it would make sense to array_unique the ids at the start of findMany, and then build a hashmap of documents and in the end loop over the original ids list to re-create the collection in the right order and with duplicates if ids have been repeated. also, getOrCreateDocuments will kill duplicates because $documents is a hashmap with the node id being the key...

dbu: oh, actually. if i read getOrCreateDocuments correctly, we return a hashmap anyways, so to keep the same behaviour as before, we can simply array_unique in findMany, as duplicates already are written to the same array field and therefore we only get each document once anyways, right?

esserj: I also think adding array_unique to `findMany` is indeed the solution we can provide for doctrine/phpcr-odm.

I do not think we can return duplicates if duplicates are asked as that would break the behaviour used for `getDocumentsByPhpcrQuery`. I'll make the PR and target both branches including a test

dbu: great, thanks!

yeah returning duplicates would also be multiple BC breaks for findMany, because a) it did not return duplicates until now and b) the index are the node ids, and we would need to change what ids we use to even allow returning duplicates :-O

tests/Doctrine/Tests/ODM/PHPCR/Functional/Translation/DocumentManagerTest.php
```diff
@@ -881,4 +883,30 @@ public function testAdditionalFindCallsDoNotRefresh()
 
         $this->assertEquals('Guten tag', $trans->topic);
     }
+
+    public function testMangerIsNotDroppingTranslationWhenDuplicateIdsArePassedToFindMany()
+    {
+        $articles = array('someText', 'moreText', 'someMoreText');
+        $article = array();
```
dbu: this line is unused as it is overwritten in the foreach immediately

esserj:
Would you be able to release this soon in at least the 1.4 branch? This way we don't have to rely on remapping to out branch.

Your time is much appreciated!

dbu:
i tagged 1.4.5 with this fix. thanks for looking into it.

if you have time to fix the travis-ci setup of the master branch, that would be amazing - it seems we got quite outdated here.

PR #796, opened 2019-10-16
fix translations not loading with duplicate ids

esserj:
it would load all none duplicates correctly and all
duplicates would load correct first time arround but
the second time the document was loaded translations
would be skipped and the new document would override
the previous document in the document collection
resulting in missing translations


dbu:
merged 1.4 to master in https://github.com/doctrine/phpcr-odm/pull/797

PR #811, opened 2020-01-24
fix children getting removed on move when pre update listener is present

esserj:
Hi @dbu ,

I discovered a critical bug that has plagued our application and it took me a while to figure out what was happening, the hardest part was trying to reproduce it in a test because I could see it going wrong, just didn't know why in out app and not in my test case. It needed multiple conditions to be true before it happened:

- fetch a parent node and iterate through its children
- update one of these children their nodename
- have a preUpdate listener present (as uow only triggers another changeset in this case)

in this case in runts through changeset 2se the first time everything goes as expected, and then updates the parents collection with the new child name and marks it moved (a move is also scheduled), but then the second time (after preUpdate) changeset is requested and the change was already applied and this it assumed the child has to be removed

### DIFF
--- a/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
+++ b/lib/Doctrine/ODM/PHPCR/UnitOfWork.php
@@ -54,6 +54,7 @@ use PHPCR\UnsupportedRepositoryOperationException;
 use PHPCR\Util\NodeHelper;
 use PHPCR\Util\PathHelper;
 use PHPCR\Util\UUIDHelper;
+use function spl_object_hash;
 
 /**
  * Unit of work class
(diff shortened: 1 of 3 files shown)


esserj:
any chance someone can have a look at this?  It's forcing us to use a patched version as there is no workaround.

lib/Doctrine/ODM/PHPCR/UnitOfWork.php
```diff
@@ -1437,8 +1438,13 @@ private function computeChildrenChanges($document, $class, $oid, $isNew, $change
                 // check moved children to not accidentally remove a child that simply moved away.
                 if (!(in_array($childName, $childNames) || in_array($childName, $movedChildNames))) {
```
dbu: does the second condition still make sense, or is it covered by checking in scheduled moves?

esserj: Nope, its not covered, the move is scheduled afterwards, so both have to exist for now, I think thats why it was detected like this initially

esserj:
@dbu thanks for taking time to review and merge!

PR #812, opened 2020-06-28
Address deprecations from persistence

greg0ire:
A backwards-compatibility layer has been added to persistence to help
consumers move to the new namespacing. It is based on class aliases,
which means the type declaration changes should not be a BC-break: types
are the same.
See doctrine/persistence#71

This means:
- using the new namespaces
- adding autoload calls for new types to types that may be extended and
use persistence types in type declarations of non-constructor methods,
so that signature compatibility is recognized by old versions of php.
More details on this at
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf


lib/Doctrine/ODM/PHPCR/Mapping/Driver/XmlDriver.php
```diff
@@ -19,10 +19,10 @@
 
 namespace Doctrine\ODM\PHPCR\Mapping\Driver;
 
-use Doctrine\Common\Persistence\Mapping\ClassMetadata;
-use Doctrine\Common\Persistence\Mapping\Driver\FileDriver;
 use Doctrine\Common\Persistence\Mapping\MappingException as DoctrineMappingException;
```
SenseException, line 22: This one looks suspicious like a candidate for a namespace change.

greg0ire: Oh right I somehow missed it

lib/Doctrine/ODM/PHPCR/Configuration.php
```diff
@@ -383,3 +385,5 @@ public function getUuidGenerator()
             };
     }
 }
+
+interface_exists(MappingDriver::class);
```
dbu, line 389: is this line on purpose?

dbu:
ah, sorry, read your blogpost now. then it makes sense. and you have the explanation in the commit message.

i assume this can still go wrong in rather confusing ways if a user manages to load the old interface before loading phpcr-odm classes that load the aliasing... 

can you please add something to the changelog for this? mention that we now support the separate persistence library, and to check the alias thing when you get the error from using the old interface and loading that before the alias gets defined (maybe with an example error message so people can recognize it)

dbu:
actually, would it not be safest if the doctrine-common files that define those interfaces look if they see the persistence interface and then define themselves as alias instead of loading their own definition? instead of every using library having to be careful about it?

dbu:
i fixed the styleci issues in 1.5 branch (and master) - can you rebase this branch please?

lib/Doctrine/ODM/PHPCR/Configuration.php
greg0ire: Yes, see the linked blog post

greg0ire:
> i assume this can still go wrong in rather confusing ways if a user manages to load the old interface before loading phpcr-odm classes that load the aliasing...

No, because the old file is precisely where the aliasing happens.

> actually, would it not be safest if the doctrine-common files that define those interfaces look if they see the persistence interface and then define themselves as alias instead of loading their own definition? instead of every using library having to be careful about it?

Autoloading everything by default means we can't have the deprecation messages, because we can no longer tell if the user tried to autoload the old type themselves.

lib/Doctrine/ODM/PHPCR/Configuration.php
dbu: sorry, i read the link only afterward and commented on my comment, but forgot to close this.

dbu:
ah, i see. yeah BC and shared libraries are really tough to get right...

thanks a lot for the effort!

lib/Doctrine/ODM/PHPCR/Proxy/ProxyFactory.php
```diff
@@ -29,6 +28,7 @@
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\ODM\PHPCR\DocumentManagerInterface;
 use Doctrine\ODM\PHPCR\Mapping\ClassMetadata;
+use Doctrine\Persistence\Mapping\ClassMetadata as BaseClassMetadata;
```
dbu, line 31: i wonder if this really works with php < 7.4.

we use this class in the skipClass method and that method is defined by the base class with the common persistence class metadata. phpstorm tells me that restricting the class will only work from php 7.4 on

greg0ire: I don't understand what you mean by "restricting the class", can you please elaborate?

greg0ire:
Next up: adding support for persistence 2 and common 3.

dbu:
we seem to also use the deprecated common proxy thing.

master is the 2.0 branch - i suggest that we could focus on upgrades there, it would also allow for BC breaks if necessary, which could make the upgrading a lot easier and cleaner.

lib/Doctrine/ODM/PHPCR/Proxy/ProxyFactory.php
dbu: this factory extends a base class that declares the skipClass method but with Doctrine\Common\Persistence\Mapping\ClassMetadata as argument.

i thought PHP 7.4 allows this, but according to https://3v4l.org/rCbc9 all PHP 7 emit a warning, while PHP 8 will fatal on this.

greg0ire:
I made a PR to 1.5, since I can't see any BC-breaks in this : #816

PR #816, opened 2020-06-30
Allow doctrine/common 3 doctrine/persistence 2

greg0ire:


dbu:
thank you!

greg0ire:
> do you plan to also update for the deprecated Common\Proxy namespace? (as said, if its easier it could also go into master only)

I wasn't planning to, no, I'm just doing this to unlock common and persistence upgrades in Sonata (plus this seems a bit more touchy). I plan to make another PR to drop persistence 1 and remove the `_exists` artefacts I added to the codebase once this is released and people have an upgrade path.

dbu:
thanks!

tell me when you need a patch release. for dropping persistence 1 and the cleanup, i guess we should go with a new minor version, right?

greg0ire:
I'd like a patch version now please :)

And yes the persistence 1 drop should go with a new minor

dbu:
i just tagged 1.5.2

PR #823, opened 2021-12-06
Extra fixes to make SF6 PR pass checks

acrobat:


.github/workflows/test-application.yaml
```diff
@@ -65,15 +65,7 @@ jobs:
...
-                  tools: 'composer:v2'
-
-            - name: Globally install symfony/flex
-              if: matrix.symfony-require != ''
-              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex
-
-            - name: Remove dev dependencies not compatible with Symfony 6
-              if: matrix.symfony-require == '6.0.*'
-              run: composer remove liip/rmt --dev --no-update --no-interaction
+                  tools: 'composer:v2, flex'
```
dbu, line 68: should we have the `,` or not? here you have one, below you don't

dbu (changes requested):
thanks a lot for the cleanups and for finding the problem with that assertion!

can you please check whether the tools value should be comma separated or just spaces?

.github/workflows/test-application.yaml
acrobat: Good catch, was copy paste error. It should indeed be seperated by a comma! Fixed!

dbu: yep, previously it only installed flex for dbal

dbu (approved):

PR #826, opened 2021-12-07
test with PHP 8.1

dbu:
seems we have some issues to fix :-/

i don't understand the pattern of failures. with php 8.0 (and 7.1-4, as per the builds on the 1.x branch) all is fine. but with php 8.1 we get errors. we do *not* get any errors with the jackrabbit storage, only with the jackalope-doctrine-dbal storage.

when running only one the ReferenceTest that has 1 error when the whole test suite is run, it has 2 errors. using phpunit 9.5, i see a different error pattern than with the currently allowed 8.5. and when phpunit 9.5 tests only the ReferenceTest, no more errors happen. however, the failures seem to be consistently happening when running the tests multiple times.

see this test run: https://github.com/doctrine/phpcr-odm/actions/runs/2979822085

anybody got some ideas?


.github/workflows/debug.yaml
```diff
@@ -0,0 +1,105 @@
+name: Debug tests
```
dbu: temporary file to focus on the failing php 8.1 build

.github/workflows/test-application.yaml
```diff
@@ -1,7 +1,7 @@
 name: Test application
 
 on:
-    pull_request:
+#    pull_request:
```
dbu: disabled until the 8.1 is problem is solved

PR #837, opened 2023-02-21
test with jackalope 2

dbu:


lib/Doctrine/ODM/PHPCR/ChildrenCollection.php
```diff
@@ -70,7 +70,7 @@ private function getNode($fetchDepth)
     {
         if (null === $this->node) {
             $path = $this->dm->getUnitOfWork()->getDocumentId($this->document);
-            $this->node = $this->dm->getPhpcrSession()->getNode($path, $fetchDepth);
+            $this->node = $this->dm->getPhpcrSession()->getNode($path, is_int($fetchDepth) ? $fetchDepth : -1);
```
SenseException, line 73: Is this a bug fix?

SenseException (approved):

lib/Doctrine/ODM/PHPCR/ChildrenCollection.php
dbu: i would say so, yes. the type declaration with strict type checking revealed that we pass `null` to phpcr, which we should not be doing according to the phpdoc types - but so far it was not enforced so we did not notice.

PR #854, opened 2023-12-03
1 to 2

dbu:


l-you:
@dbu Any update on this PR?
Perhaps I could help with something?

dbu:
thanks for asking. i wanted to remove annotations completely in https://github.com/doctrine/phpcr-odm/pull/858 rather than trying to fix legacy things for the upcoming version. but did not find time to continue on that. it looks like not much is missing to wrap that one up - if you want to give it a go you would be very welcome, otherwise i will try again next week.

PR #858, opened 2023-12-04
remove obsolete annotation mapping

dbu:


dbu:
will ignore the resource report by phpstan, seems to be a bug: https://github.com/phpstan/phpstan/issues/10399

PR #863, opened 2024-01-12
split find into find with required class name and findDocument for finding any document

### LINKED ISSUE #862: ObjectManager::find requires class name
new versions of doctrine/persistence force the $className argument of ObjectManager::find to be string.

phpcr-odm allows to find without specifying the class. we need to add a new method for this use case that does not have a className argument at all in Doctrine\ODM\PHPCR\DocumentManagerInterface and update the documentation

dbu:
This is a BC break, but one necessary to accomodate future base doctrine versions as the className in ObjectManager::find becomes typed.

fix #862


dbu:
fixed in #864 without BC breaks.

PR #865, opened 2024-01-15
modernize php code

dbu:


l-you:
@dbu Great job! Thank you so much!
Which stages remain to release phpcr-odm 2.0 ?

dbu:
honestly mostly stage fright :D 

i just tagged https://github.com/doctrine/phpcr-odm/releases/tag/2.0.0-beta2

if you have an application with phpcr-odm, it would be great if you can try the beta2 and report on any issues. we seem to never have started an UPGRADE.md but at least the changelog should explain a bit what needs to be done. if there are BC breaks that are missing, i am also glad if you can report those.

llupa:
@dbu hey! 

> honestly mostly stage fright :D 

Totally get this, but since this is the internet 😄 do you have any _feeling_ when would you move this forward? 
I am pinging because of this: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1933

dbu:
we are currently analyzing the last couple of regressions with sulu, one of the heavy users of the phpcr ecosystem. i hope that by the end of the month, i can tag a stable release.

have you been able to build FOSElasticaBundle with the latest dev-2.x version? that would be a good check to know if we break anything for you.

PR #867, opened 2024-04-02
fix cs to avoid deprecation in php 8.3

dbu:


lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterBase.php
```diff
@@ -117,7 +117,7 @@ protected function dispatchMany(array $nodes): void
      *
      * @return object|array - PHPCR QOM object or array of objects
      */
-    public function dispatch(AbstractNode $node)
+    public function dispatch(QBConstants $node)
```
SenseException: `AbstractNode` seems to be in the same namespace. Neither `use` is needed nor the alias.

lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterPhpcr.php
```diff
@@ -290,7 +290,7 @@ protected function walkOperandStaticLiteral(OperandStaticLiteral $node): Literal
     {
         $value = $node->getValue();
 
-        if ($field = $node->getParent()?->getChildOfType(AbstractNode::NT_OPERAND_DYNAMIC)) {
+        if ($field = $node->getParent()?->getChildOfType(QBConstants::NT_OPERAND_DYNAMIC)) {
```
SenseException, line 293: `AbstractNode` seems to be in the same namespace. Neither use is needed nor the alias.

lib/Doctrine/ODM/PHPCR/Query/Builder/QueryBuilder.php
```diff
@@ -28,7 +28,7 @@
  *
  * @author Daniel Leech <daniel@dantleech.com>
  */
-class QueryBuilder extends AbstractNode
+class QueryBuilder extends QBConstants
```
SenseException, line 31: Same here. Did you chose to use an alias for a reason? A renaming in a future major version maybe?

SenseException (changes requested):

lib/Doctrine/ODM/PHPCR/Query/Builder/ConverterBase.php
dbu: that was the cs fixer. i think someone originally aliased AbstractNode to QBConstants for using the constants, while keeping AbstractNode as typehint... changed it all to AbstractNode now.

PR #875, opened 2025-08-16
Feature: add support for ClassLocator from doctrine/persistence 4.1

rela589n:
The changes integrate a forward compatibility support for https://github.com/doctrine/persistence/pull/433 
`ClassLocator`, which allows clients to pass any iterable of classes they might want.

Parallel PRs for:
- ODM: https://github.com/doctrine/mongodb-odm/pull/2802
- ORM: https://github.com/doctrine/orm/pull/12131

Note that it doesn't provide full support for persistence 4.x, but instead implements `AttributeDriver`, while also adjusting some signatures to achieve compatibility with 4.x.

### DIFF
--- a/.phpstan/phpstan-baseline.neon
+++ b/.phpstan/phpstan-baseline.neon
@@ -40,3 +40,24 @@ parameters:
             message: "#^Call to an undefined method Doctrine\\\\ODM\\\\PHPCR\\\\Query\\\\Builder\\\\AbstractNode\\:\\:setConverter\\(\\)\\.$#"
             count: 1
             path: ../tests/Doctrine/Tests/ODM/PHPCR/Query/Builder/QueryBuilderTest.php
+
+        -
+            message: '#^Access to an undefined property Doctrine\\ODM\\PHPCR\\Mapping\\Driver\\AttributeDriver\:\:\$classLocator\.$#'
+            identifier: property.notFound
+            count: 1
+            path: ../lib/Doctrine/ODM/PHPCR/Mapping/Driver/AttributeDriver.php
+            reportUnmatched: false
+
+        -
+            message: '#^Class Doctrine\\Persistence\\Mapping\\Driver\\ClassLocator not found\.$#'
+            identifier: class.notFound
+            count: 1
+            path: ../lib/Doctrine/ODM/PHPCR/Mapping/Driver/AttributeDriver.php
+            reportUnmatched: false
+
+        -
+            message: '#^Parameter \$paths of method Doctrine\\ODM\\PHPCR\\Mapping\\Driver\\AttributeDriver\:\:__construct\(\) has invalid type Doctrine\\Persistence\\Mapping\\Driver\\ClassLocator\.$#'
+            identifier: class.notFound
+            count: 2
+            path: ../lib/Doctrine/ODM/PHPCR/Mapping/Driver/AttributeDriver.php
+            reportUnmatched: false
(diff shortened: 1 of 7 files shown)


.phpstan/phpstan-baseline.neon
```diff
@@ -40,3 +40,24 @@ parameters:
             message: "#^Call to an undefined method Doctrine\\\\ODM\\\\PHPCR\\\\Query\\\\Builder\\\\AbstractNode\\:\\:setConverter\\(\\)\\.$#"
             count: 1
             path: ../tests/Doctrine/Tests/ODM/PHPCR/Query/Builder/QueryBuilderTest.php
+
+        -
+            message: '#^Access to an undefined property Doctrine\\ODM\\PHPCR\\Mapping\\Driver\\AttributeDriver\:\:\$classLocator\.$#'
+            identifier: property.notFound
+            count: 1
+            path: ../lib/Doctrine/ODM/PHPCR/Mapping/Driver/AttributeDriver.php
+            reportUnmatched: false
```
dbu, line 49: why do we need these in the baseline? can't we make sure that phpstan runs with the latest version of persistence?

dbu (commented):
thanks a lot for looking into this. i will have a look at the github actions, i guess we are using a too old docker image to run the tests.

dbu:
i fixed the CI setup, please rebase this branch on 2.0.x

i updated phpstan to run on php 8.4, maybe that solves the issue with the persistance false alerts?

.phpstan/phpstan-baseline.neon
rela589n: The problem is that a complete upgrade to `doctrine/persistence 4.0` requires many more changes. The current code is made to be forward-compatible with `ColocatedMappingDriver` of persistence 4.1, but the actual upgrade to this version has yet to be implemented.

rela589n: What I did locally is:

1. I adjusted a "require" section:

```json
"doctrine/data-fixtures": "^1.0 || ^2.0",
"doctrine/persistence": "^3.0 || ^4.0@dev",
```

2. Ran `AttributeDriverTest`.

At first, the tests were failing due to incompatible method / property declarations.

Those that could be easily fixed w/o breaking the current tests on persistence 3.x, I've committed.
Those that would break the tests on persistence 3.x, I've only monkey-patched locally to verify that `AttributeDriverTest` works.

Therefore, current tests run on persistence 3.x.

dbu: ah, i see. so as is, these changes can not be used, its only preparation for persistence 4 compatibility?

dbu:
tests seem green now. should i merge so that we have this part of the work to upgrade? for full compatibility, further work is needed.

.phpstan/phpstan-baseline.neon
rela589n: These changes will kick in once persistence 4 comes into play

rela589n:
I'm wondering which version this should be merged into, since the full upgrade toward 4.0 persistence would require BC breaks. Actually, these changes to the signatures themselves are BC breaks.

dbu:
> I'm wondering which version this should be merged into, since the full upgrade toward 4.0 persistence would require BC breaks. Actually, these changes to the signatures themselves are BC breaks.

oh indeed. i can create a 3.x branch if you want.

rela589n:
> create a 3.x branch if you want

It would be more correct, yet it's all up to you

dbu:
i created the 3.0.x branch. can you edit the target of this PR?

my idea for supporting older versions (of the base doctrine libs) is: as long as its only very small effort and does benefit users (e.g. keep supporting older PHP versions) we should. but if it becomes an obstacle, either a lot of effort, or ugly solutions, or not fully profiting from the new features of the base libs, we should ditch old versions.

dbu:
hi @rela589n are you still on this? should i merge the current state or do you want to do more things in this branch?

rela589n:
Hi, @dbu , I think you can merge it, since full support for 4 would imply much work which I'm not willing to dedicate to. Current changes provide the first steps to it.

dbu (approved):
thanks a lot for the contribution

dbu:
i created https://github.com/doctrine/phpcr-odm/milestone/6 and a new issue there so we don't forget to do that before we release phpcr-odm 3.

PR #879, opened 2025-11-24
Add support for Symfony 8

loic425:


Rafikooo (approved):

loic425:
There are issues with doctrine persistence 4...

dbu:
thanks for looking into this loic!

> There are issues with doctrine persistence 4...

yep, seems signatures changed. do we have to allow persistence 4 for symfony 8 support? if we do, we might need some ugly BC hack, or think if we can at the same time remove support for persistence 3 and ourselves bump a minor version.

loic425:
@dbu Indeed, Symfony 8 is only a dev requirement on doctrine persistence. So it should be ok.

loic425:
Hum.... For now, at least doctrine-bundle 3.1.x branch supports Symfony 8 https://packagist.org/packages/doctrine/doctrine-bundle#3.1.x-dev, but this branch requires at least doctrine persistence 4.

dbu:
blocked by https://github.com/liip/RMT/pull/185

dbu:
i added php 8.5 builds and fixed deprecations. now the only remaining thing is probably RMT

dbu (approved):
and with RMT updated, this passes :tada:

loic425:
that's great, but we need to use persistence 4 to use with the Doctrine bundle.

dbu:
lets see https://github.com/doctrine/phpcr-odm/pull/881

PR #881, opened 2025-12-04
allow installation with doctrine persistence 4

### LINKED ISSUE #878: Support doctrine/persistence 4
support for ClassLocator and some preliminary work has been done in #875, but more adjustments will be needed.

if its no extra effort, we can keep supporting older versions of persistence, but if thats not easily doable, bump to persistence 4.1.

dbu:
* [x] target 3.x branch
* [x] adjust to issues shown by phpstan
* [x] adjust DocumentManagerInterface to better macht EntityManagerInterface (array instead of Collection)
* [ ] ~adjust ProxyFactory to make use of PHP native lazy objects (look at https://github.com/doctrine/orm/blob/3.6.x/src/Proxy/ProxyFactory.php)~ => #883 
* [x] check if there are other deprecated things

fix #878

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -59,7 +59,7 @@ jobs:
 
         steps:
             - name: Checkout project
-              uses: actions/checkout@v5
+              uses: actions/checkout@v6
 
             - name: Install and configure PHP
               uses: shivammathur/setup-php@v2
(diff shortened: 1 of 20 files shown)


dbu:
uh. this will need a new major version. a couple of BC breaks on the interfaces are necessary.

loic425:
Ok, take your time, but indeed, a new major version is necessary. Fo the packages I maintain, ODM packages are just optional, so I don't need this quickly.

dbu:
@rela589n i created this to see what all fails. there is quite some things. if you feel up to it, separate pull requests to continue aligning the phpcr codebase are very appreciated.

so far one significant issue seems to be that persistence returns `array` on findAll / findMany, while we returned a Collection. this only shows now because of strict typing, it was a mismatch before that too, from what i can tell.

rela589n:
Hi, @dbu !

> issue that persistence returns array on findAll / findMany

Yeah, that's the problem

